PR #26, opened 2022-02-11
Corrections of surface velocity + minor changes

facusapienza21:
Corrected the way in which the surface velocity is calculated. Instead of calculate the surface velocity, the previous version of the code was computing the vertical average velocity. The surface velocity can be found in Equation (8.35) in K.M. Cuffey and W.S.B. Paterson. 2010, The physics of glaciers, Fourth edition.


JordiBolibar (approved):
Good point! I remember this different formulation between average vertical vs surface velocities when I was exploring the SIA. This will allow a direct comparison with remote sensing observations.

PR #85, opened 2022-12-02
Batching working for ice rheology inversions

JordiBolibar:
First version of batching working. The loss function and the training still needs exploration to become efficient.


JordiBolibar:
@facusapienza21 You can play with this. The training still needs some experimentation. BFGS seems to be pretty slow, so if you get stuck with that, try using ADAM.

PR #111, opened 2023-05-29
Continuous integration for tests 

facusapienza21:
Based on the workflow already existing in the @JordiBolibar [fork](https://github.com/JordiBolibar/ODINN.jl), I managed to make the test to run when creating the conda environment and later running the code in Julia. 

## The solution

The problem was coming from the creation of the conda environment itself. For some reason I am not sure, the environment wasn't properly been set before. I transition to `micromamba` in order to create the environment, which seems to be doing the job but also it is very fast. I also notice the original `environment.yml` file was overspecified, so I worked in creating a very minimalistic `environment.yml` that installs the required dependencies. Notice that OGGM and their appendices () are installed using `pip`. 

Once the environment has been created, every action in the workflow will execute the `oggm_env` as long as the shell is properly initialized. By default, if we use `shell: bash -el {0}` this will initialized the shell with the right environment (and this is how we can run Julia with the right environment). Notice I added this as the default shell in the preamble. 

I also added these changes in a new workflow called `test.yml`. I think is a good idea to have the different actions in independents files, so we can run multiple actions depending the task. 

## Comments

Right now I am using `OGGM=1.6`. I originally did this because I was trying to run some tests using the last OGGM API for testing. Furthemore, we are currently using a developer mode version of OGGM in our current workflow (`1.5.4.dev53+gfd6ce6c`) which is not directly available though pip. However, notice that the [MBSandbox](https://github.com/OGGM/massbalance-sandbox) does not work with the last version of OGGM (`1.6`), so I am not sure if this will lead to future problems at the moment of actually running the test. 

Notice that in order to make the action to work we may need to change the permission to execute workflows coming form PR. In `Setting > Actions > General` to allow all actions. 

## Warning ⚠️⚠️⚠️

At the moment of accepting the pull requests, instead of accept and merge, we have to do `Confirm squash and merge` in order to all the commit messages in my history to be compressed to a single one.


JordiBolibar:
I had a hard time trying to merge this with my fork, so I ended up doing a manual merge for this into my branch dedicated to fixing the CI. These changes are now available in #116, so I'm closing this one.

PR #170, opened 2025-01-23
Migrate to pure julia

albangossard:


Project.toml
```diff
@@ -50,21 +49,20 @@ ChainRules = "1.50"
...
-OptimizationOptimJL = "0.1.5"
+OptimizationOptimJL = "0.3"
 OptimizationOptimisers = "0.1.2, 0.2"
 OrdinaryDiffEq = "5, 6"
 PlotThemes = "3.1"
 Plots = "1.2"
 Polynomials = "2, 3, 4"
 ProgressMeter = "1"
-PyCall = "1.9"
 PyPlot = "2.11"
```
JordiBolibar, line 66: This should be removed as well

Project.toml
```diff
@@ -32,7 +32,6 @@ PlotThemes = "ccf2f8ad-2431-5c83-bf29-c5338b663b6a"
 Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 Polynomials = "f27b6e38-b328-58d1-80ce-0feddd5e7a45"
 ProgressMeter = "92933f4c-e287-5a05-a399-4b506db050ca"
-PyCall = "438e738f-606a-5dbb-bf0a-cddfbfd45ab0"
 PyPlot = "d330b81b-6aea-500a-939a-2ce795aea3ee"
```
JordiBolibar, line 35: Same.

scripts/new_API_tests.jl
JordiBolibar, line 1: Hmm, I think this name is not very informative. We should perhaps rename it, since from now one the new API will be the default one, so no need to name it this way. 

I'm not even entirely sure this test is needed. It seems to be testing a forward run (i.e. Huginn stuff) directly from ODINN. If it's redundant we might just remove it for good.

scripts/toy_model.jl
JordiBolibar, line 1: I'm not sure this file will be valid with the new architecture. This was just a script to showcase the old one for the paper. If it cannot be run with the new architecture we could just remove it.

src/helpers/mass_balance.jl
JordiBolibar, line 1: OK, now I understand why you said there were still Python code here. This is some legacy stuff that should be removed. I don't know why it's still here. This is now taken up by Sleipnir.jl, so you can remove it.

src/helpers/utils.jl
JordiBolibar, line 1: Same for this. Actually all the `helpers` folder should be gone. Now the helpers are spread out among packages.

src/parameters/Hyperparameters.jl
```diff
@@ -14,7 +14,7 @@ end
     Hyperparameters(;
         current_epoch::Int64 = nothing,
         current_minibatch::Int64 = nothing,
-        loss_history::Vector{Float64} = Vector{Float64}[],
+        loss_history::Vector{Float64} = zeros(Float64, 0),
```
JordiBolibar, line 17: Why do you want to initialize it as zeros? I think the idea was to have an empty array and to append new loss values as the training advanced.

src/parameters/UDEparameters.jl
```diff
@@ -13,7 +13,7 @@ Base.:(==)(a::UDEparameters, b::UDEparameters) = a.sensealg == b.sensealg && a.o
 
 """
     UDEparameters(;
-        sensealg::SciMLBase.AbstractAdjointSensitivityAlgorithm = InterpolatingAdjoint(autojacvec=ReverseDiffVJP(true)),
+        sensealg::SciMLBase.AbstractAdjointSensitivityAlgorithm = GaussAdjoint(autojacvec=EnzymeVJP()),
```
JordiBolibar, line 16: Cool, let's leave it like this for now and see what @ChrisRackauckas finds out to be the best solution.

JordiBolibar (changes requested):
Just a few changes to double check. I think there's still some legacy stuff in `helpers` that can be completely removed.

Project.toml
albangossard: Oups you're right thank you, corrected in a4586d2

Project.toml
albangossard: Thanks!
a4586d2

scripts/new_API_tests.jl
albangossard: Yes this test is redundant with what is done in Huginn. This file was deleted in a4586d2.

scripts/toy_model.jl
albangossard: Removed in a4586d2

src/helpers/mass_balance.jl
albangossard: a4586d2

src/helpers/utils.jl
albangossard: Thanks for the precision. The `helpers` folder was deleted in a4586d2.

albangossard (commented):
Thanks for the review! Following your comments, I've cleaned several files to remove the legacy code.

JordiBolibar (approved):
OK great, looks good! We can move on and continue the AD/adjoint fixes in ODINN.jl 😄

Issue #152, opened 2024-05-28
Fully migrate from `Flux.jl` to `Lux.jl`


JordiBolibar:
Follow the migration to Lux.jl from #151, we should remove legacy code from Flux.jl related to the `DataLoader`. Something similar exists with MLutils.jl.

facusapienza21:
I migrated all the Flux code to Lux in #229

PR #229, opened 2025-03-01
Migrate all `Flux` to `Lux`

facusapienza21:
This PR includes a transition of all the Flux code to Lux. Flux is no longer a dependency of ODINN here. 

This addesses #152 

@albangossard please notice that this PR includes your commits from #228. I did this so I can work in changing Flux to Lux with your type stable code. This does not solve yet the problem, but we may get closer. You should merge your PR to `main` before we merge the branch `fix-adjoints` to `main`, but except the lines with `Enzyme.API.runtimeActivity!(true)` I haven't made significant changes.


facusapienza21:
Currently tests are not passing since Huginn has Optimization libraries among their dependencies. In order to run in local, just run Huginn in development mode without the optimization libraries or from this [fork](https://github.com/facusapienza21/Huginn.jl).

PR #231, opened 2025-03-03
Fix various issues including type stability

albangossard:
Hey @facusapienza21 @JordiBolibar, now we have the inversion unit test that is running with the type stable code. The UDE unit test is still buggy. There is a segmentation fault during Enzyme compilation. I have managed to extract the traceback which was previously hidden behind other errors (I removed the pmap):

<pre><font color="#26A269"><b>julia&gt; </b></font>ude_solve_test(atol; MB=true)
<font color="#2AA1B3"><b>[ Info: </b></font>Filtering out these glaciers from RGI ID list: Any[]
Running training of UDE...

<font color="#A2734C"><b>┌ Warning: </b></font>Number of observations less than batch-size, decreasing the batch-size to 1
<font color="#A2734C"><b>└ </b></font><font color="#5E5C64">@ MLUtils ~/.julia/packages/MLUtils/EDvou/src/batchview.jl:104</font>
Training iceflow UDE...

[111138] signal (11.1): Segmentation fault
in expression starting at REPL[24]:1
CanonicalizeInPlace at /workspace/srcdir/Enzyme/enzyme/Enzyme/TypeAnalysis/TypeTree.h:767
updateAnalysis at /workspace/srcdir/Enzyme/enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp:1091
considerTBAA at /workspace/srcdir/Enzyme/enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp:1237
analyzeFunction at /workspace/srcdir/Enzyme/enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp:6001
CreateFromClone at /workspace/srcdir/Enzyme/enzyme/Enzyme/GradientUtils.cpp:4351
CreateAugmentedPrimal at /workspace/srcdir/Enzyme/enzyme/Enzyme/EnzymeLogic.cpp:2290
EnzymeCreateAugmentedPrimal at /workspace/srcdir/Enzyme/enzyme/Enzyme/CApi.cpp:669
EnzymeCreateAugmentedPrimal at /home/gossarda/.julia/packages/Enzyme/QsaeA/src/api.jl:404
unknown function (ip: 0x71670c394c18)
_jl_invoke at /cache/build/tester-amdci4-11/julialang/julia-master/src/gf.c:2895 [inlined]
ijl_apply_generic at /cache/build/tester-amdci4-11/julialang/julia-master/src/gf.c:3077
[...]

PR #232, opened 2025-03-03
Small changes to find error in fixTypeStability

facusapienza21:
@albangossard I created this small PR since I was trying to make your PR #231 to work. As I mentioned, the dummy gradient should work. A few weird things that I noticed here: 
- For some reason, the `inn1()` inside the computation of the loss function does not seem to be necesary anymore. Do you know why? I couldn't find any change affecting this. 
- It seems that `Enzyme.Const` really crashes stuff, even the dummy gradient. This is the reason I am not sure the segmentation fault you saw is coming from the pure AD problem, but of course I can be wrong. 

Now you will see that the inverse simulation with dummy grad works, but there is a problem: the parameter \theta does not get updated... this didn't happen before (notice one of the test actually checks there are updates in the parameter vector). Do you think some of the changes may have broke this?

Happy to keep debugging this together! I feel every PR bring us closer to a final solution!!!

### DIFF
--- a/src/simulations/functional_inversions/functional_inversion_utils.jl
+++ b/src/simulations/functional_inversions/functional_inversion_utils.jl
@@ -105,7 +105,8 @@ function loss_iceflow(θ, batch_ids::Vector{I}, simulation::FunctionalInversion)
             end
             # Ice surface velocities
             normVref = mean(Vx_ref.^2 .+ Vy_ref.^2)^0.5
-            l_V_loc = loss(Vx_pred, inn1(Vx_ref)) + loss(Vy_pred, inn1(Vy_ref))
+            # l_V_loc = loss(Vx_pred, inn1(Vx_ref)) + loss(Vy_pred, inn1(Vy_ref))
+            l_V_loc = loss(Vx_pred, Vx_ref) + loss(Vy_pred, Vy_ref)
             l_V += normVref^(-1) * l_V_loc
         else
             # Ice thickness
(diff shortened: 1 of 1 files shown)


albangossard:
Hey @facusapienza21, just a few comments. I'll have a look at your changes on tomorrow morning.

> For some reason, the inn1() inside the computation of the loss function does not seem to be necesary anymore. Do you know why? I couldn't find any change affecting this.

I experienced the same problem. For an unknown reason at some point the `inn1` function wasn't necessary anymore. I commented it and then later on I needed it again, so I'm not sure whether we need to remove it.

> It seems that Enzyme.Const really crashes stuff, even the dummy gradient. This is the reason I am not sure the segmentation fault you saw is coming from the pure AD problem, but of course I can be wrong.

As discussed on slack, I don't think the segmentation fault is because of the dummy gradient. The same error occurred on the unit test of the UDE. This might because of `Enzyme.Const` but I'm not 100% sure either as I needed that transformation in order for the unit test of the inversion to work (obviously it's a different line of code but this shows that sometimes we need this transformation).

I'm glade to see that we're moving forward!

src/simulations/functional_inversions/functional_inversion_utils.jl
```diff
@@ -153,7 +154,8 @@ function batch_iceflow_UDE(θ, simulation::FunctionalInversion, batch_id::I) whe
 
     tstops = Huginn.define_callback_steps(params.simulation.tspan, params.solver.step)
     params.solver.tstops = tstops
-    stop_condition(u,t,integrator) = Sleipnir.stop_condition_tstops(u,t,integrator, Enzyme.Const(tstops)) #closure
+    # stop_condition(u,t,integrator) = Sleipnir.stop_condition_tstops(u,t,integrator, Enzyme.Const(tstops)) #closure
+    stop_condition(u,t,integrator) = Sleipnir.stop_condition_tstops(u,t,integrator, tstops)
```
albangossard, line 158: From today's debugging, these `Enzyme.Const` calls are not the root cause of the segmentation fault but ok to remove them for the moment if it makes the dummy gradient tests work!

albangossard (approved):
This looks good to me. Let's merge this so that you can continue working on the adjoint.

As for the weights that don't get updated, the only reason I could identify is that we moved from Flux to Lux and the way these weights are defined as changed. A simple test would be to revert to Flux and see if this solves that issue.

PR #239, opened 2025-03-19
Custom adjoint testing

ChrisRackauckas:
This PR is to track the custom adjoint branch and Enzyme compatibility.


src/inverse/gradient.jl
```diff
@@ -0,0 +1,348 @@
...
+    dθs = ODINN.merge_batches(dθs)
+
+    if maximum(norm.(dθs)) > 1e7
+        @show losses
+        @show norm.(dθs)
+        @warn "Potential unstable gradient for glacier"
+    end
+
+    @assert typeof(θ) == typeof(sum(dθs))
+    dθ = sum(dθs)
```
JordiBolibar, line 41: We found a typo here. This should be `dθ .= sum(dθs)`.

JordiBolibar:
I did some debugging and gained some insight on where the error might be coming from. I didn't push anything since we still need to integrate several changes to this branch. I added the necessary pieces of code to debug and reproduce what I found.

JordiBolibar:
We have added manual implementations of the Discrete adjoint (working) and Continuous adjoint (still not validated), which can be used as well as a ground truth. 

To debug the latest version of the code using `MWE_custom_adjoint.jl`, you can pass `EnzymeAdjoint` as the type of `UDEparameters.grad`. The VJP in `gradient.jl` using Enzyme for `∂f∂H` is correct, but the one for `∂f∂θ` is not.

facusapienza21:
@JordiBolibar @albangossard let's move the discussion here regarding merging our recent changes into `main`. It seems that once we have #269 merged into `customAdjoint` and after doing releases of the other libraries (so we are sure all is working with CI) we can move forward to have the latest architecture into main.

PR #241, opened 2025-03-24
Custom adjoint tests

albangossard:
Implementation of the tests to check the adjoint of the ODE with both the continuous and the discrete adjoints.

For the moment the tests of the adjoint of SIA2D are spread between Huginn and ODINN. I think that in the future we should move the implementation of `VJP_λ_∂SIA∂H_continuous` and `VJP_λ_∂SIA∂θ_continuous` to Huginn as this will allow us to check the alignment of the forward and backward in the same repository and simplify the development process (the tests in `ODINN/test/SIA2D_adjoint.jl` will move to Huginn). Note that since we need the neural network in `VJP_λ_∂SIA∂θ_continuous`, it isn't clear how to do that.

With the current state of the code, only the `test_grad_discreteAdjoint()` test passes as there seems to be an issue in the implementation of the continuous adjoint.


test/grad_free_test.jl
```diff
@@ -28,7 +28,7 @@ function grad_free_test(;use_MB::Bool=false)
                                                 optimizer=ODINN.ADAM(0.01)),
                         UDE = UDEparameters(sensealg=sensealg, 
                                             optim_autoAD=adtype, 
-                                            grad=DummyAdjoint(grad=dummy_grad), 
+                                            grad=ODINNDummyAdjoint(grad=dummy_grad),
```
JordiBolibar, line 31: Why this name? Why not keep `DummyAdjoint`? If we want to have ODINN on it it can already be `ODINN.DummyAdjoint`.

test/test_grad_loss.jl
```diff
@@ -0,0 +1,233 @@
...
+        machine_learning = NeuralNetwork(params))
+
+    # We retrieve some glaciers for the simulation
+    glaciers = initialize_glaciers(rgi_ids, params)
+
+    # Time stanpshots for transient inversion
+    tstops = collect(2010:δt:2015)
+
+
+    ### Fake law for A(T) for Peterson & Cuffey
```
JordiBolibar, line 52: Since this part is common to both tests, maybe we could externalize it from the functions?

JordiBolibar (changes requested):
Good job! Nice progress towards having an architecture that supports multiple types of adjoint. I have just added a few minor comments. 

We can try to merge it with the `customAdjoint` branch, but the `gradient.jl` file might be problematic. Once we have merged both PRs we can try that.

src/inverse/AdjointTypes.jl
```diff
@@ -6,11 +6,13 @@ export ODINNEnzymeAdjoint, ODINNZygoteAdjoint
 abstract type AbstractAdjointMethod end
 
 @kwdef struct ODINNContinuousAdjoint{F <: AbstractFloat} <: AbstractAdjointMethod
+    # Continuous adjoint of SIA2D with manual implementation of the backward in the ODE scheme
```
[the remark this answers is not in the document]
albangossard: You're right thank you, filled docstrings of this file in e6a1756

test/grad_free_test.jl
albangossard: Just because `DummyAdjoint` was renamed `ODINNDummyAdjoint` at some point and I have updated this name so that this script runs without error.

PR #240, opened 2025-03-24
[WIP] Debugging Enzyme VJPs ongoing

JordiBolibar:
Enzyme seems to run, but we need to correctly validate the gradient of the VJP of dfdH. The gradient of dfdtheta is still zero, so something is wrong and needs to be fixed.

PR #242, opened 2025-03-25
Custom continuous adjoint + tests of the different adjoints

albangossard:
This PR also integrates the changes of https://github.com/ODINN-SciML/ODINN.jl/pull/240

### DIFF
--- a/src/ODINN.jl
+++ b/src/ODINN.jl
@@ -68,6 +68,7 @@ const global root_plots::String = joinpath(root_dir, "plots")
 
 include(joinpath(root_dir, "src/setup/config.jl"))
 # All parameters needed for the models
+include(joinpath(root_dir, "src/inverse/AdjointTypes.jl"))
 include(joinpath(root_dir, "src/parameters/Hyperparameters.jl"))
 include(joinpath(root_dir, "src/parameters/UDEparameters.jl"))
 # Simulations
@@ -77,6 +78,7 @@ include(joinpath(root_dir, "src/simulations/inversions/Inversion.jl"))
 # ML models
 include(joinpath(root_dir, "src/models/machine_learning/MLmodel.jl"))
 # Inversion 
+include(joinpath(root_dir, "src/inverse/SIA2D_adjoint.jl"))
 include(joinpath(root_dir, "src/inverse/gradient.jl"))
 
 end # module
--- a/src/inverse/AdjointTypes.jl
+++ b/src/inverse/AdjointTypes.jl
@@ -1,0 +1,27 @@
+export AbstractAdjointMethod
+export ODINNContinuousAdjoint, ODINNDiscreteAdjoint
+export ODINNDummyAdjoint
+export ODINNEnzymeAdjoint, ODINNZygoteAdjoint
+
+abstract type AbstractAdjointMethod end
+
+@kwdef struct ODINNContinuousAdjoint{F <: AbstractFloat} <: AbstractAdjointMethod
+    reltol::F = 1e-6
+    abstol::F = 1e-6
+end
+
+@kwdef struct ODINNDiscreteAdjoint{F <: AbstractFloat} <: AbstractAdjointMethod
+    step::F = 1/12
+end
+
+@kwdef struct ODINNDummyAdjoint <: AbstractAdjointMethod
+    grad::Function
+end
+
+@kwdef struct ODINNEnzymeAdjoint{F <: AbstractFloat} <: AbstractAdjointMethod
+    step::F = 1/12
+end
+
+@kwdef struct ODINNZygoteAdjoint{F <: AbstractFloat} <: AbstractAdjointMethod
+    step::F = 1/12
+end
--- a/src/inverse/AdjointTypes.jl
+++ b/src/inverse/AdjointTypes.jl
@@ -1,27 +1,81 @@
 export AbstractAdjointMethod
-export ODINNContinuousAdjoint, ODINNDiscreteAdjoint
-export ODINNDummyAdjoint
-export ODINNEnzymeAdjoint, ODINNZygoteAdjoint
+export ContinuousAdjoint, DiscreteAdjoint
+export DummyAdjoint
+export EnzymeAdjoint, ZygoteAdjoint
 
+"""
+    AbstractAdjointMethod
+
+Abstract type representing the flavor of AD and adjoint to be used to compute the
+gradient of the cost function. There are two parts where one can play with how the
+gradient is propagated: the iceflow model VJP and the adjoint of the ODE solver.
+The VJP of the iceflow model can be computed using either AD (Zygote or Enzyme), the
+discrete, or the continuous adjoint of the iceflow model.
+As for the computation of the adjoint of the ODE solution, it can be handled by
+SciMLSensitivity, or computed using the adjoint implemented in ODINN.
+"""
 abstract type AbstractAdjointMethod end
 
-@kwdef struct ODINNContinuousAdjoint{F <: AbstractFloat} <: AbstractAdjointMethod
+"""
+Struct that defines the SciMLSensitivity adjoint flavor. This is the default
+behavior in ODINN.
+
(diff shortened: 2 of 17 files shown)


src/inverse/AdjointTypes.jl
```diff
@@ -0,0 +1,27 @@
+export AbstractAdjointMethod
+export ODINNContinuousAdjoint, ODINNDiscreteAdjoint
```
JordiBolibar, line 2: I'm not convinced by these names, they are unnecessarily verbose. We're already in ODINN, so I don't think we need to specify it in each type.

src/inverse/SIA2D_adjoint.jl
```diff
@@ -0,0 +1,150 @@
+export VJP_λ_∂SIA∂H_continuous
+export VJP_λ_∂SIA∂θ_continuous
+"""
+Implementation of continuous adjoint to time-independent SIA2D equation.
```
JordiBolibar, line 4: Here I'd also boost a bit the docstrings using GitHub Copilot :)

src/inverse/gradient.jl
```diff
@@ -82,97 +87,99 @@ function SIA2D_grad_batch!(θ, simulation::FunctionalInversion)
...
 
-            # Zygote adjoint implementation
-            # Create pullback function to evaluate VJPs
-            # dH_H, ∂f∂H_pullback = Zygote._pullback(∂f∂H_closure, H[j])
+            # β = 2.0
+            # normalization = std(H_ref[j][H_ref[j] .> 0.0])^β
+            normalization = 1.0
+            # Compute derivative of local contribution to loss function
+            # TODO: Update this based on the actual value of the loss function as ∂(parameters.UDE.empirical_loss_function)/∂H
+            ∂ℓ∂H = 2 .* (H[j] .- H_ref[j]) ./ (prod(N) * Δx * Δy * normalization)
```
JordiBolibar, line 105: If I remember correctly from a previous PR, I think you removed the `Δx * Δy` part?

src/inverse/gradient.jl
```diff
@@ -82,97 +87,99 @@ function SIA2D_grad_batch!(θ, simulation::FunctionalInversion)
...
+            ℓ += Δt[j-1] * simulation.parameters.UDE.empirical_loss_function(H[j], H_ref[j]) / normalization
 
-            # Compute VJP with adjoint variable
-            # Transpose operation
-            # _, λ_∂f∂H = ∂f∂H_pullback(λ[j])
-            # λ_∂f∂H, = ∂f∂H_pullback(λ[j])
+            if typeof(simulation.parameters.UDE.grad) <: ODINNZygoteAdjoint
```
JordiBolibar, line 108: We can keep this with `if` statements for now, but this looks like a perfect example to use multiple dispatch. We could just wrap all the workflows in a function with multiple methods for each adjoint type. It would be super clean :)

src/inverse/gradient.jl
JordiBolibar, line 1: All code from lines 219 to 308 are just duplicated functions. I already removed them from the latest version of my PR. I think I'd wait until my PR is merged, because it will highly impact this PR.

JordiBolibar (changes requested):
A few changes. I would still wait for #240 to be merged, since it will highly impact this PR. I removed quite a lot of redundant code in `gradient.jl`.

albangossard:
> A few changes. I would still wait for https://github.com/ODINN-SciML/ODINN.jl/pull/240 to be merged, since it will highly impact this PR. I removed quite a lot of redundant code in gradient.jl.

I just have incorporated your changes in `customAdjointContinuous`, so merging the current PR will automatically incorporate the changes you made in #240. I needed your changes to handle the conflicts and align the adjoint with the refactoring you did in Huginn. Doing that way allows us to avoid having to resolve the merge conflict twice. Are you ok with that?

JordiBolibar:
> > A few changes. I would still wait for #240 to be merged, since it will highly impact this PR. I removed quite a lot of redundant code in gradient.jl.
> 
> I just have incorporated your changes in `customAdjointContinuous`, so merging the current PR will automatically incorporate the changes you made in #240. I needed your changes to handle the conflicts and align the adjoint with the refactoring you did in Huginn. Doing that way allows us to avoid having to resolve the merge conflict twice. Are you ok with that?

Yes, OK. I see the tests are not passing, it seems Enzyme cannot be precompiled. I would first fix that and then we can merge :)

src/inverse/gradient.jl
albangossard: Yes this term should be removed but the PR you're referring to is https://github.com/ODINN-SciML/ODINN.jl/pull/241 and it hasn't been merged into [customAdjointContinuous](https://github.com/ODINN-SciML/ODINN.jl/tree/customAdjointContinuous) yet.

src/inverse/gradient.jl
albangossard: Very good idea! I created an issue for that https://github.com/ODINN-SciML/ODINN.jl/issues/244

src/inverse/gradient.jl
albangossard: I'm not sure which lines you're referring to since the line numbers have changed with the new commits. Your changes should have been incorporated by now.

src/inverse/AdjointTypes.jl
albangossard: Yes I agree with that. These variables were renamed in 92c6bfc. This will be integrated in the current PR once https://github.com/ODINN-SciML/ODINN.jl/pull/241 is merged.

src/inverse/SIA2D_adjoint.jl
albangossard: Done in 92c6bfc :)

albangossard:
Thanks for the review!

> Yes, OK. I see the tests are not passing, it seems Enzyme cannot be precompiled. I would first fix that and then we can merge :)

I'm working on that, I'll ping you once everything is ready to be merged

albangossard:
@facusapienza21 in case you want a summary of the changes we did by integrating the commits related to the adjoint to the ones related to Enzyme, you can take a look at this PR

src/ODINN.jl
```diff
@@ -77,6 +78,7 @@ include(joinpath(root_dir, "src/simulations/inversions/Inversion.jl"))
 # ML models
 include(joinpath(root_dir, "src/models/machine_learning/MLmodel.jl"))
 # Inversion 
+include(joinpath(root_dir, "src/inverse/SIA2D_adjoint.jl"))
```
JordiBolibar, line 81: Not super important, but I think it might be better to include this within `SIA2D.jl` rather than at root level, since it should be included "as a pack".

src/models/machine_learning/ML_utils.jl
```diff
@@ -116,13 +120,22 @@ function generate_plot_folders(path)
...
-const A_values = hcat(A_values_sec[1,:], A_values_sec[2,:].*60.0*60.0*24.0*365.25)'
+"""
+    A_law_PetersonCuffey()
 
-# Polynomial fit for Cuffey and Paterson data 
-A_f = Polynomials.fit(A_values[1,:], A_values[2,:]) # degree = length(xs) - 1
+Returns a law of the coefficient A as a polynomial of the temperature.
+The values used to fit the polynomial come from Peterson & Cuffey.
+"""
+function A_law_PetersonCuffey()
```
JordiBolibar, line 129: This should be Paterson :)

JordiBolibar (approved):
Two really minor comments, otherwise it's all good! Feel free to merge if you just want to fix those later :)

src/ODINN.jl
albangossard: Unless I'm mistaken, it's not possible because `SIA2D.jl` is within Huginn but the two files we need to import here are within ODINN.

src/models/machine_learning/ML_utils.jl
albangossard: Good catch! Fixed in b658f8e

PR #266, opened 2025-04-02
Fixes in Continuous adjoint and MWE working for multiple glaciers

facusapienza21:


src/inverse/gradient.jl
```diff
@@ -194,20 +193,32 @@ function SIA2D_grad_batch!(θ, simulation::FunctionalInversion)
...
-                λ[j-1] .= λ[j] .+ Δt[j-1] .* (λ_∂f∂H .- ∂ℓ∂H)
+                λ[j-1] .= λ[j] .+ Δt[j-1] .* (λ_∂f∂H .+ ∂ℓ∂H)
 
                 λ_∂f∂A = Huginn.SIA2D_discrete_adjoint(λ[j-1], H[j], simulation, t₀; batch_id = i)[2]
                 ∇θ, = Zygote.gradient(_θ -> grad_apply_UDE_parametrization(_θ, simulation, i), θ)
                 λ_∂f∂θ = λ_∂f∂A*∇θ
 
-                dLdθ .+= - Δt[j-1] .* λ_∂f∂θ # The minus is needed here, not clear why
+                ### Update gradient
+                dLdθ .+= Δt[j-1] .* λ_∂f∂θ
```
albangossard, line 203: We also need to update this formula in the Zygote and Enzyme if blocks. Maybe we could mutualize that line just before `push!(dLdθs_vector, dLdθ)` to avoid any inconsistency

src/models/machine_learning/ML_utils.jl
```diff
@@ -13,18 +13,17 @@ Generates a neural network.
...
-    #     Dense(10, 3, x -> softplus.(x)),
-    #     # Dense(3, 1, sigmoid)
-    #     Dense(3, 1, sigmoid_A)
+    #     Dense(3, 1, Base.Fix2(sigmoid_A, lims))
     # )
+    UA = Lux.Chain(
+        Dense(1, 3, x -> softplus.(x)),
+        Dense(3, 10, x -> softplus.(x)),
+        Dense(10, 3, x -> softplus.(x)),
+        Dense(3, 1, Base.Fix2(sigmoid_A, lims))
```
albangossard, line 25: This `Base.Fix2` function makes Enzyme fail. I think we should find another way to provide the bounds.
`ERROR: Enzyme execution failed.
Enzyme: active by ref type Active{Base.Fix2{typeof(ODINN.sigmoid_A), Tuple{Float64, Float64}}} is overwritten in application of custom rule for MethodInstance for LuxLib.Impl.fused_dense!(::Matrix{Float64}, ::LuxLib.LoopedArrayOp, ::Base.Fix2{typeof(ODINN.sigmoid_A), Tuple{Float64, Float64}}, ::Base.ReshapedArray{Float64, 2, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}, Tuple{}}, ::Matrix{Float64}, ::SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}) val=  %40 = getelementptr inbounds { { [2 x double] }, i64, i64 }, { { [2 x double] }, i64, i64 } addrspace(11)* %0, i64 0, i32 0, !dbg !384 ptr=  %68 = getelementptr inbounds [1 x { [2 x double] }], [1 x { [2 x double] }] addrspace(11)* %67, i64 0, i32 0, !dbg !447. As a workaround until support for this is added, try passing values as separate arguments rather than as an aggregate of type Active{Base.Fix2{typeof(ODINN.sigmoid_A), Tuple{Float64, Float64}}}.`

albangossard (changes requested):
Thank you @facusapienza21 for these fixes!
I couldn't run the code as there are several features that are currently broken. Enzyme fails at differentiating the `Base.Fix2` function and the continuous adjoint test of SIA2D has errors much larger than the threshold.
I'll try to debug that today and if I can find a fix, I'll commit it.

src/inverse/gradient.jl
```diff
@@ -26,16 +26,16 @@ function SIA2D_grad!(dθ, θ, simulation::FunctionalInversion)
...
     dθs  = getindex.(loss_grad, 2)
     dθs = ODINN.merge_batches(dθs)
 
     if maximum(norm.(dθs)) > 1e7
-        @show losses
-        @show norm.(dθs)
-        @warn "Potential unstable gradient for glacier"
+        glacier_ids = findall(>(1e7), norm.(dθs))
+        for id in glacier_ids
+            @warn "Potential unstable gradient for glacier $(simulation.glaciers[id].rgi_id): ‖dθ‖=$(norm.(dθs)[id]) \n Try reducing the temporal stepsize Δt used for reverse simulation."
```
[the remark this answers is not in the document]
albangossard: I made the change in 0bb670b, feel free to correct me if the initial implementation was the one you wanted

test/SIA2D_adjoint.jl
```diff
@@ -87,17 +87,19 @@ function test_adjoint_SIAD2D_continuous()
...
         println("Gradient wrt H")
         println("eps    = ",printVecScientific(eps))
         println("ratio  = ",printVecScientific(ratio))
         println("angle  = ",printVecScientific(angle))
         println("relerr = ",printVecScientific(relerr))
     end
-    @test (min_ratio<thres_ratio) & (min_angle<thres_angle) & (min_relerr<thres_relerr)
+    @test min_ratio<thres_ratio
+    @test min_angle<thres_angle
+    @test min_relerr<thres_relerr
```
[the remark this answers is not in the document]
albangossard: Fixed in the recent commits

src/inverse/gradient.jl
albangossard: Done in 0bb670b

src/models/machine_learning/ML_utils.jl
albangossard: I removed that function and the bounds are now provided in a different way. There are still errors with Enzyme but I'll fix them later.

albangossard:
@facusapienza21 I made some changes to this branch. The alignment of the continuous adjoint is much better now. We dropped the error from 60% to 2%. Let me know what you think of these changes and if you think they are correct.

src/models/machine_learning/ML_utils.jl
facusapienza21: Sounds good! Enzyme was crashing in my computer anyways and probably was because of this (although I don't remember seeing this in the error message). The other way of passing the bounds for the sigmoid works just as good!

facusapienza21:
Could be a good idea to address #267 here!

facusapienza21:
@albangossard  I made several changes in the last commit: 

- Notice now what we called discrete and continuous are both under discrete adjoint and I created a new separated type for the continuous adjoint that computes the time-integrated adjoint and gradient using the ODE solvers (and then automatically handling the time-stepping method to ensure stability)
- Created a new struct/type for the VJP type. I think this is more clear, since we have on one side the time-integrated adjoint and then how do we compute VJPs (same logic as in SciMLSensitivity but for our custom adjoints now). 
- MWE example now working with larger values of A
- Test passing with the new continuous adjoint and discrete VJP. The discrete adjoint with cont VJP still does not pass the test (and it may never do it, since technically this is computing a different gradient). 

All indicates that this is working (both tests and MWE) but please let me know if you see something off. I will probably keep working on this tomorrow, so don't feel pressure to merge this PR immediately (but do in case you in the Grenoble side are thinking or making some conflicting changes in `customAdjoint`).

facusapienza21:
Btw, I think in the future we may want to change the names cont and discrete adjoints to be more accurate (as written now, the discrete is just a subcase of the continuous). This is not a big deal for now and we can keep it as it is, but good to keep in mind! Happy to discuss more about it.

albangossard:
Thank you @facusapienza21 for this huge step up in the functionality of the custom adjoint. Having the continuous in time adjoint is so nice.

I agree that the previous names of the AD flavors were misleading and I'm 100% aligned with how you named these new flavors. From now one, I think that we should refer to the backward of SIA2D as `EnzymeVJP`, `ContinuousVJP` or `DiscreteVJP` as you did, and I would recommend that we don't use the term "adjoint" when we're are only speaking of differentiating the SIA2D function. In opposition, the terms "discrete adjoint" and "continuous adjoint" make clear that we're referring to the differentiation of the ODE problem. What is your opinion about that?
Cc @JordiBolibar in case you want to take part in the discussion.

I played with the continuous adjoint and I think that it's working well. The values or the error in the unit tests suggest that we're computing the good gradient.

I made some changes and the unit tests are now passing in one single shot. There is some obscure bug when we try to perform an inversion before the unit test of SIA2D with Enzyme, so I just changed the order of the tests. I also addressed https://github.com/ODINN-SciML/ODINN.jl/issues/267

As the tests are passing, the CI should also work, and I think it's a good time to merge the different `customAdjoint` branch and make new releases. That way, we can continue to work with a reliable CI that pings us when we break something. What do you think? @facusapienza21 @JordiBolibar

albangossard:
@facusapienza21 In 8cfb5e1 I created a VJP functions that use multiple dispatch to select which code should be selected to compute the VJPs. This makes the manual backpropagation of the gradient agnostic to which VJP is used as everything is handled in `ODINN/src/inverse/AD_utils.jl`. The other advantage is that the code is mutualized and you can seamlessly switch between DiscreteVJP and EnzymeVJP in your implementation of the continuous adjoint!

facusapienza21:
Oh I love the last commit @albangossard ! Very nice! 

Let me just take one more look at this today and maybe we can merge tomorrow (feel free to merge on your Friday unless there is some open matter to address). 

I fully agree that we should not longer refer to the VJP computation inside SIA as adjoint (altougth it is also an adjoint) since it may be confessing. I think the new naming makes all the sense for the continuous adjoint (cont in time which controls stability, and then discrete/cont/enzyme on space for the VJPs). I am not fully satisfied with the name of DiscreteAdjoint since this is not technically the discrete adjoint (it should use the computational steps of the forward model with the same time-steps, which may take a bit longer to implement since we need to store info of the forward pass).

facusapienza21:
Notice that CI is not passing since we haven't release yet the last version of Sleipnir with the changes in the branch `customAdjoint` (in Sleipnir).

albangossard:
> Oh I love the last commit @albangossard ! Very nice!

Yeah the multiple dispatch makes this so elegant.

> Let me just take one more look at this today and maybe we can merge tomorrow (feel free to merge on your Friday unless there is some open matter to address).

No problem I'll wait your approval to merge this.

> I am not fully satisfied with the name of DiscreteAdjoint since this is not technically the discrete adjoint

Yes you're right, and to be precise and avoid any confusion we should rename it. What do you think of `ConstantTimeSteppingAdjoint`? Once we've implemented the true discrete adjoint, we'll remove this flavor, but in the meantime we can rename it.

> Notice that CI is not passing since we haven't release yet the last version of Sleipnir with the changes in the branch customAdjoint (in Sleipnir).

That's why I would like to make the different releases. I think that we've converged to a satisfying point and we the other packages shouldn't change that much given that the next step is to handle multiprocessing, and which hopefully should only impact ODINN.

facusapienza21:
@albangossard @JordiBolibar the new version of the code now works with both ADAM and BFGS combined. Take a look at how the two optimizers are handle and let me know what you think. This code was use to generate the following simulation with 8 glaciers using 50 epochs of ADAM and 50 of LBFGS, without multiprocessing, which took ~1.5hrs in my local computer. 
image

albangossard (approved):
These results are promising @facusapienza21!
How you handle the optimizers look good to me.

JordiBolibar:
I have not been able to catch up with everything here. Lots of exciting changes. Super happy to see that the training is finally working smoothly, and congrats @albangossard on the implementation of multiple adjoints with multiple dispatch, that's neat! Don't wait for my review on this one, merge and move on with the rest of PRs (including mine).

PR #268, opened 2025-04-04
Add VJP benchmark in CI

albangossard:
Hi @facusapienza21! I added a benchmark that will allow us to compare which changes affect the performance of the code. The idea is to be able to compare the performance between different commits and between our different AD flavors. We can run the benchmark using `benchmark/run.jl`, but it will be launched by the CI anyway as soon as a PR is opened. The results are then posted in the PR discussion with a bot.
I also implemented that for the other packages. You can have a look at the Sleipnir and Muninn PRs to see how it's formatted.
The PRs associated to each package are the following:
- https://github.com/ODINN-SciML/Sleipnir.jl/pull/94
- https://github.com/ODINN-SciML/Muninn.jl/pull/23
- https://github.com/ODINN-SciML/Huginn.jl/pull/74


facusapienza21:
Oh nice! I was just wondering how to do this yesterday, so great you did it!

PR #269, opened 2025-04-04
Multiprocessing and precompilation improved

JordiBolibar:
I have improved a few things in terms of precompilation, making sure we only rely on the necessary dependencies. Moreover, I have re-implemented the forward run in the inversion for the ground truth to fully run in parallel, and I have verified that the training runs in parallel. There is also a (commented) implementation of package compilation, which is way too slow to be useful, but which we might be interested in fixing in the future.


src/models/machine_learning/ML_utils.jl
```diff
@@ -182,16 +182,50 @@ Arguments:
...
-    T = mean(glacier.climate.longterm_temps)
-    A = fakeA(T)
     # Generate a fake forward model for the simulation
-    generate_glacier_prediction!(glacier, params, model; A = A, tstops=tstops)
+    if law == :PatersonCuffey
+        A_poly = ODINN.A_law_PatersonCuffey()
+        fakeA(T) = A_poly(T)
+    else
+        @error "Unknown law of A: $law"
+    end
```
albangossard, line 197: This implementation is great!
Maybe we can keep the option to provide a polynomial, don't you think? It might be useful to generate fake data based on a dummy polynomial without having to implement it in ODINN.

src/models/machine_learning/ML_utils.jl
```diff
@@ -182,16 +182,50 @@ Arguments:
...
+        ts = prediction.results[i].t
+        Hs = prediction.results[i].H
+    
+        @assert ts ≈ tstops "Timestops of simulated PDE solution and UDE solution do not match."
+
+        if isnothing(prediction.glaciers[i].data)
+            prediction.glaciers[i].data = [Sleipnir.ThicknessData(ts, Hs)]
+        else
+            append!(prediction.glaciers[i].data, Sleipnir.ThicknessData(ts, Hs))
+        end
```
albangossard, line 227: One honest question: isn't it better in terms of performance to initialize `prediction.glaciers[i].data` as a list before the loop? I guess this avoid the compiler to generate if/else statements in the compiled code which can improve the performance. Here it shouldn't matter that much but that's a point I'm frequently asking myself.
I guess it might also improve type stability.

albangossard (approved):
It should have been so painful to try to make it work!
I have only one minor comment and one honest question. Otherwise the remaining looks good to me

src/ODINN.jl
```diff
@@ -59,6 +57,9 @@ cd(@__DIR__)
 const global root_dir::String = dirname(Base.current_project())
 const global root_plots::String = joinpath(root_dir, "plots")
 
+const SYSIMAGE_DIR = joinpath(homedir(), ".ODINN")
+const SYSIMAGE_PATH = joinpath(SYSIMAGE_DIR, "odinn_sysimage.so")
+ENV["JULIA_DEPOT_PATH"] = joinpath(homedir(), ".julia")  # Ensure shared cache
```
facusapienza21, line 62: @JordiBolibar what is this needed for?

facusapienza21 (commented):
This looks all good to me! I would just be sure that this changes now integrate well with the latest PR on customAdjoint before merging.

src/ODINN.jl
JordiBolibar: This is just for PackageCompiler.jl to create a system image. I have so far commented that part of the code, so I'll also comment this line.

src/models/machine_learning/ML_utils.jl
JordiBolibar: I have added this option. We should nonetheless further complexify this, since right now the law is "hardcoded", mapping only `T` to `A`. In the future we'll come up with more complex parametrizations, so we'll have to figure out a clean and flexible way to have multiple mappings.

src/models/machine_learning/ML_utils.jl
JordiBolibar: I'm not sure I'm following your question. I'm not seeing what it refers to in the code.

JordiBolibar:
> This looks all good to me! I would just be sure that this changes now integrate well with the latest PR on customAdjoint before merging.

I resolve the conflicts. We should double check that everything looks good, and that the tests which so far pass are working.

JordiBolibar:
@facusapienza21 I have pushed the last WIP commit. As mentioned, there seems to be some tests which are failing. The first ones seem to be off by a small margin. Interestingly, the tests are not running with multiprocessing, so I'm not sure the numerical differences come from the different workers. If you have some time please take a look, otherwise I'll continue tomorrow.

facusapienza21:
@JordiBolibar I am not sure what you say is not working: I see all the tests passing in CI, except for the last one where the problem does not seems to be numerical.

JordiBolibar:
Interesting, it was not working in local for me. If that's the case feel
free to merge.

El dt., 8 d’abr. 2025, 20:27, Facundo Sapienza ***@***.***>
va escriure:

PR #274, opened 2025-04-08
Merge changes from `customAdjoint`

facusapienza21:
@JordiBolibar @albangossard I created this branch based on customAdjoint so we can merge the changes without altering the PR https://github.com/ODINN-SciML/ODINN.jl/pull/239 open with Chris as suggested by Jordi.

I just notice I misspelled adjoint in this PR: another good reason to merge this PR soon.


facusapienza21:
@albangossard do you think we can change the long message about "computing finite differences" in the CI? I think it is ok to print it, but right now there are many being printed in the log!

albangossard:
I didn't find a quick way to fix the lines that are printed many times, so I just disabled the print of the progress bar in the CI: https://github.com/ODINN-SciML/Huginn.jl/pull/78
It will still be printed when you execute from your local computer.

JordiBolibar:
OK, I have merged and fixed some issues with the PR for the documentation. Now everything should be good. I'll wait for the tests to pass and then we can merge into main! :)

PR #294, opened 2025-04-09
Ice velocity integration

albangossard:
The purpose of this PR is to integrate ice velocity into the loss for transient inversions.
Associated PR:
https://github.com/ODINN-SciML/Huginn.jl/pull/79
https://github.com/ODINN-SciML/Sleipnir.jl/pull/97


facusapienza21:
Wow I love this new feature of a draft pull request @albangossard ! I will start using it!

albangossard:
In order for the CI to pass, Sleipnir and Huginn need to be released with the changes from the PRs mentioned above. However, today I started to have Enzyme issues on my local computer. The tests that rely on Enzyme are failing with a very long stack trace. This used to work and I think that a dependency has changed. We should revert to the proper configuration, and eventually open an issue if we can isolate a minimal reproducible example.

JordiBolibar (approved):
Seems all good to me. We can merge when we solve those Enzyme issues.

Issue #275, opened 2025-04-09
Discrete vs Continuous VJP: accuracy and performance


facusapienza21:
I would like to understand better the differences between the discrete and continuous VJP methods to compute the gradient. Ideally, I want to campare them more exhaustedly (no urgent), but I am creating this issue to keep track of this. In principle, both methods give the following accuracies when compared to finite differences:

***Discrete VJP*** 
```bash
> Testing adjoint ContinuousAdjoint{Float64, Int64, DiscreteVJP}(DiscreteVJP(), RDPK3Sp35{typeof(OrdinaryDiffEqCore.trivial_limiter!), typeof(OrdinaryDiffEqCore.trivial_limiter!), Static.False}(OrdinaryDiffEqCore.trivial_limiter!, OrdinaryDiffEqCore.trivial_limiter!, static(false)), 1.0e-8, 1.0e-8, 0.08333333333333333, 200)
[ Info: Running forward in-place PDE ice flow model
Processing glacier RGI60-11.03638 for PDE forward simulation
Computing gradient of θ using finite differences with ϵ=1.0e-03... 100%|
eps    =  1.00e-03  1.00e-04  1.00e-05  1.00e-06  1.00e-07  1.00e-08
ratio  = -5.21e-04  2.08e-04 -4.35e-05  5.11e-04 -5.56e-03  9.35e-02 
angle  = -4.03e-08 -4.86e-09 -8.35e-07 -9.63e-06 -4.15e-04 -1.08e-02 
relerr =  5.94e-04  2.30e-04  1.29e-03  4.42e-03  2.94e-02  1.64e-01 
```

***Continuous VJP***
```bash
> Testing adjoint ContinuousAdjoint{Float64, Int64, ContinuousVJP}(ContinuousVJP(), RDPK3Sp35{typeof(OrdinaryDiffEqCore.trivial_limiter!), typeof(OrdinaryDiffEqCore.trivial_limiter!), Static.False}(OrdinaryDiffEqCore.trivial_limiter!, OrdinaryDiffEqCore.trivial_limiter!, static(false)), 1.0e-8, 1.0e-8, 0.08333333333333333, 200)
eps    =  1.00e-03  1.00e-04  1.00e-05  1.00e-06  1.00e-07  1.00e-08
ratio  =  3.31e-02  3.39e-02  3.36e-02  3.42e-02  2.79e-02  1.30e-01 
[...]

PR #296, opened 2025-04-10
Increase distance_to_border in halfar tests to reduce the error

albangossard:
Please see https://github.com/ODINN-SciML/ODINN.jl/issues/275

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -44,25 +44,25 @@ ENV["GKSwstype"]="nul"
 
 @testset "Parameters constructors with specified values" params_constructor_specified()
 
-@testset "Consistency between discrete adjoint and Enzyme AD" test_grad_Enzyme_SIAD2D() # This test must be run first, otherwise Enzyme compilation fails because it was used before
+@testset "Discrete VJP vs Enzyme VJP" test_grad_Enzyme_SIAD2D() # This test must be run first, otherwise Enzyme compilation fails because it was used before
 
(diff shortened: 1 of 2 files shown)


test/runtests.jl
```diff
@@ -60,9 +60,9 @@ ENV["GKSwstype"]="nul"
 
 @testset "Manual backward of the loss terms vs Enzyme" test_grad_loss_term()
 
-@testset "Manual implementation of the discrete VJP vs Enzyme for Halfar solution" test_grad_Halfar(ContinuousAdjoint(VJP_method=DiscreteVJP()); thres=[5e-1, 1e-15, 5e-1])
+@testset "Manual implementation of the discrete VJP vs Enzyme for Halfar solution" test_grad_Halfar(ContinuousAdjoint(VJP_method=DiscreteVJP()); thres=[3e-1, 1e-15, 5e-1])
```
facusapienza21, line 63: Following the naming of the test, can I suggest calling this "Implementation of continuous adjoint with discrete VJP vs Enzyme for Halfar solution"?.  I will also remove all the "manuals" from the tests, but that is another story.

facusapienza21 (commented):
Looks good! Just a little comment.

test/runtests.jl
albangossard: Done in 1f57545 ;)

facusapienza21:
All looks good! I will proceed to merge!

Issue #295, opened 2025-04-09
Move continuous adjoint and discrete VJP code together


facusapienza21:
Right now we have the continuous VJP method implemented in ODINN and the discrete VJP in Huginn. I think all should be in the same space. I think it makes more sense to have all in ODINN since Huginn is just for forward simulation, but I think both options are fine. 

What do you think @albangossard ? I can take care of this, but just want to check with you what do you think before moving forward!

facusapienza21:
Following my comment in Slack, I think it makes more sence to move things to ODINN (and not Huginn) since the type of inversion will determine some internals of the VJPs, which right now live in Huginn. 

By doing this, I will also move the tests of the VJPs from Huginn to ODINN. 

Let me know if you approve this @albangossard and then I will move forward with a PR in the following days!

albangossard:
Yes we should definitely standardize things!

I understand your opinion about moving the VJPs to ODINN but I think that from a software development point of view, it's much better to move the VJPs to Huginn for two reasons:
- Everything related to the iceflow should be in Huginn and this includes the VJPs although they are not really used in this package.
- If we move the VJPs to ODINN and we make some changes in the SIA2D, we won't see that we break the consistency between the forward and the VJP in the CI until we make a release of Huginn.

I think that the VJPs shouldn't depend on the neural network and it should only compute VJPs with respect to the input of SIA2D, that is with respect to H, A, D, etc. This is actually what was implemented for the discrete VJP and this allows to check the consistency without needing to define a neural network. Additionally, this makes the finite difference test faster for A since we only need to differentiate a scalar.

What is your opinion about this ? I'm happy to discuss how we could standardize the way we test VJPs.

facusapienza21:
Thank you for the comments @albangossard ! I partially disagree with the first argument, but I fully agree with your second argument. **However, the VJPs will depend of the neural network.** For the simple inversion of A we where working until now, the dependency of the VJP between the physical parameters (H, A, D) and the NN (\theta) was separable, as you wrote in your code where the \nabla \theta is applied after VJPs. However, in more complicated cases this won't be longer the case. A good example is when the whole D is replaced by a NN that takes H as argument. 

[...]

PR #298, opened 2025-04-11
Design of API for custom target inversion

facusapienza21:
This (draft) PR is to track the progress in the design of custom targets, or different learning objectives. At least for SIA, we probably want to desing a model for the diffusivity (and MB maybe in the future) that combines physical components (e.g., some functional dependence with H, \nabla H, etc) and some modelling components (e.g., a NN parametrized by \theta). 

This can be done by defining a custom type `Target()` that include the required information to do this and by-pass the VJP/adjoint calculation without modifying our beautiful adjoint code. This is still work in progress, so early feedback is welcome. 

Important note: I anticipate the components of `Target()` will change as I converge to the best solution. These partial derivatives can be computed with AD (and I think this will do SciMLSensitivity the day things work) but I think this will be more efficient, since these are partial derivatives related to the ML design: the heavy work is carried by the adjoint calculation. We can discuss this @JordiBolibar to see differences.

- [x] Implement changes in Discrete VJP
- [x] Implement changes in Continuous VJP
- [x] Design of an inversion of D (not just A)


JordiBolibar:
Yesterday we had a brief discussion with @albangossard on this. This design will be key to have an intuitive API for users to "play" with the differential equations and to substitute subparts easily. 

My first idea is that most cases can be covered with the already existing `SIA2D!` function that we have. The whole "magic" should be done **outside** the function, mapping the output of the regressors to whatever variable(s) we want to parametrize in the SIA.

There are two potential usecases:
- For mapping single granular variables, we can keep the current workflow: e.g. just injecting the output of the regressor to `C` or `A`. 
- For mapping lumped variables (e.g. `D`), we can also pass them into `SIA2D!`, but we need a clever way to tell the function "this is `D`, bypass all the calculations of the diffusivity and directly inject `D` into the SIA". 

It's the second case which will require something smart to make it smooth and flexible, in order to avoid hard-coding `ifelse` statements all the way. 

I suggest we discuss this in the next weekly meeting.

facusapienza21:
I am not sure more cases will be covered with what we have in `SIA2D`: a diffusivity where we parametrized a neural network NN that takes as an input H for example. This operation needs to be distributed among pixels and then aggregated. The reason why we can make this division between NN and SIA is because the neural network does not change spatially or temporally. The temporal part can be handled in ODINN (this is what we are doing with the adjoint) but the spatial variabity is done right now in Huginn for the discrete VJPs and then the separation between SIA and NN cannot be done that straigforward.  

I think it will be easier to discuss this in person.

facusapienza21:
@JordiBolibar @albangossard this is still work in progress and I am sure there is a lot of place for improvement, but I made some changes in the code to handle the inversion as we discussed on Tuesday. I designed this so it works with the A inversion as it was before and a new D inversion. I haven't take a look of how to fix the Enzyme tests, mostly because Enzyme crashes in my computer... I can look at this at some point too. 

I was mostly working on the design of the target inversion, the re-design of how VJPs are computed in the case of a general inversion, and how new experiments can be designed using the `Target`. Fell free to comment/review or we can directly discuss this in our next meeting.

@albangossard notice I moved the VJP code from Huginn to here together with the respective tests (I also made a draft PR in Huggin with the required changes)

JordiBolibar:
Hi @facusapienza21, thanks for all this work! I'll take a look and give you feedback on this. 

As you might have seen, we spent quite a lot of time yesterday discussing the new API for the functional inversions and the laws, which we have summarized in #302. The goal should be to connect that high-level API with your `Target` internals. In my review, I will also try to add comments on how to best connect these two interfaces.

README.md
```diff
@@ -54,7 +54,7 @@ params = Parameters(simulation = SimulationParameters(working_dir=working_dir,
 		    hyper = Hyperparameters(batch_size=4,
 					    epochs=10,
 					    optimizer=ODINN.ADAM(0.01)),
-		    UDE = UDEparameters(target = "A")
+		    UDE = UDEparameters(target = :A)
```
JordiBolibar, line 57: As mentioned in #302, the idea would be to map targets with functions directly in the new `Law` types which would be fed into iceflow and mass balance models.

src/inverse/SIA2D/adjoint.jl
```diff
@@ -0,0 +1,415 @@
...
+    dSdx = Huginn.diff_x(S) ./ Δx
+    dSdy = Huginn.diff_y(S) ./ Δy
+    ∇Sx = Huginn.avg_y(dSdx)
+    ∇Sy = Huginn.avg_x(dSdy)
+
+    # Compute surface slope
+    ∇S = (∇Sx.^2 .+ ∇Sy.^2).^(1/2)
+
+    # Compute average ice thickness
+    H̄ = Huginn.avg(H)
```
JordiBolibar, line 226: I realize we should probably export `avg` to have a cleaner interface.

src/inverse/SIA2D/adjoint.jl
```diff
@@ -0,0 +1,415 @@
...
+- `λ::Matrix{R}`: Adjoint state, also called output gradient in reverse-mode AD.
+- `H::Matrix{R}`: Ice thickness which corresponds to the input state of the SIA2D.
+- `simulation::SIM`: Simulation parameters.
+- `t::R`: Time value, not used as SIA2D is time independent.
+- `batch_id::Union{Nothing, I}`: Batch index.
+
+Returns:
+- `dλ::Matrix{R}`: Jacobian vector product, also called input gradient in reverse-mode AD.
+"""
+function VJP_λ_∂SIA∂θ_continuous(
```
JordiBolibar, line 316: Just out of curiosity, but do these need to be out-of-place or they could potentially work in-place?

src/models/machine_learning/MLmodel.jl
```diff
@@ -47,11 +48,13 @@ Feed-forward neural network.
...
     ChainType <: Lux.Chain,
     ComponentVectorType <: ComponentVector,
-    NamedTupleType <: NamedTuple
+    NamedTupleType <: NamedTuple,
+    TAR <: AbstractTarget
 } <: MLmodel
     architecture::ChainType
     θ::ComponentVectorType
     st::NamedTupleType
+    target::TAR
```
JordiBolibar, line 57: We will have to think where to exactly place the `Target`, since the following #302 the main idea was to drop `MLmodel` altogether, and only have iceflow and mass balance models.

src/models/machine_learning/MLmodel.jl
```diff
@@ -65,20 +68,45 @@ Creates a new feed-forward neural network.
 - `architecture`: `Flux.Chain` neural network architecture (optional)
 - `θ`: Neural network parameters (optional)
 """
-function NeuralNetwork(params::P;
-            architecture::Union{ChainType, Nothing} = nothing,
-            θ::Union{ComponentArrayType, Nothing} = nothing) where {P <: Sleipnir.Parameters, ChainType <: Lux.Chain, ComponentArrayType <: ComponentArray}
+function NeuralNetwork(
```
JordiBolibar, line 71: I think we should also discuss if we should specify our own machine learning models, or we should completely externalize them. I think it's still a good idea to provide some "default" NNs of different sizes, since for many tasks the type of MLP will be the same. 

And as for my previous comment, in #302 we came up with a different way to "connect" the regressor with a target, inside a `Law` type. I think that perhaps your `Target` type should go into the `Law` type, basically constructing the internals of whatever high-level decision has been made through the `Law` interface.

src/models/machine_learning/Target_utils.jl
JordiBolibar, line 1: All these functions will have to be somehow connected to the logic of `Law`. There might be some redundancy in some aspects that we'll have to navigate.

src/parameters/UDEparameters.jl
```diff
@@ -10,17 +10,17 @@ A mutable struct that holds parameters for a UDE (Universal Differential Equatio
...
+mutable struct UDEparameters{ADJ<:AbstractAdjointMethod} <: AbstractParameters
     sensealg::SciMLBase.AbstractAdjointSensitivityAlgorithm
     optim_autoAD::AbstractADType
     grad::Union{ADJ, Nothing}
     optimization_method::String
     loss_type::String
     empirical_loss_function::AbstractLoss
     scale_loss::Bool
-    target::Union{String, Nothing}
+    target::Union{Symbol, Nothing}
```
JordiBolibar, line 23: Same, probably we want to move that inside the new `Law` type.

JordiBolibar (commented):
Awesome work, thanks a lot for pulling this out! My comments only go in the direction of trying to connect this with #302. @facusapienza21 @albangossard @mleseach, I'd be happy to hear your comments on this. I suggest we start the conversation here, so we have some baseline for the next meeting.

src/inverse/SIA2D/adjoint.jl
facusapienza21: Yes, worth checking of this. If we include the `avg` function, we also probably want to include all the other similar utils from Huginn. I am not fully agains to leave it as it is tough.

src/models/machine_learning/MLmodel.jl
facusapienza21: Would this means just to move ML model inside a broader model? Or do you mean to get rid of Target? Notice that target has inside all the required internals to compute the VJPs and gradient, this information needs to be provided somewhere.

src/models/machine_learning/MLmodel.jl
facusapienza21: Right now I am allowing the two options, and I think we should keep it like that: either provide the architecture and we handle the setup of the NN, or don't provide the architecture and just build the default NN. We do want to be able to design the architecture, since this is one of the tools the modeler would like to explore and we want to be as flexible as possible with this, including different architectures, activation functions, and initializations of the weights. Allowing to pass the Lux.Chain is the perfect compromise between these, which is what I currently did in one of the last commits.

src/models/machine_learning/Target_utils.jl
facusapienza21: Yes, we need to think about it. I still don't see how the vision of #302 with Law matches with this. We should discuss this in person!

facusapienza21:
@JordiBolibar @albangossard please take a look at the latest changes in the code. Any feedback is welcome! Please still don't merge into main, although I don't think we will merge this PR without a couple of review iterations.  

Resolves #295

src/models/target/target_D.jl
```diff
@@ -0,0 +1,229 @@
...
+    cv = zero(cv_template)
+    for i in 1:length(v)
+        cv[i] = v[i]
+    end
+    return cv
+end
+
+function ComponentVector2Vector(cv::ComponentVector)
+    return collect(cv)
+end
```
albangossard, line 229: Wouldn't it be more natural to move these functions to `target_utils.jl`?

src/models/target/target_D.jl
```diff
@@ -0,0 +1,229 @@
...
+
+### Normalization functions to ensure the scales of input are comparable to each other
+
+function normalize_T(T; lims)
+    return (T .- lims[1]) ./ (lims[2] - lims[1]) .- 0.5
+end
+
+function normalize_H(H; lims)
+    return (H .- lims[1]) ./ (lims[2] - lims[1]) .- 0.5
+end
```
albangossard, line 207: These two functions are the same and could be mutualized, aren't they?

src/models/target/target_D.jl
```diff
@@ -0,0 +1,229 @@
...
+        for i in axes(H, 1), j in axes(H, 2)
+            ∇θ_point, = Zygote.gradient(_θ -> predict_A_target_D(_θ, temp, H[i,j]; ml_model = ml_model, params = params), θ)
+            ∂D∂θ[i, j, :] .= ∂A_spatial[i, j] * ComponentVector2Vector(∇θ_point)
+        end
+    else
+    # elseif mode == "interp"
+        # Interpolation of the gradient as function of values of H.
+        # Introduces interpolation errors but it is faster and probably sufficient depending
+        # the decired level of precision for the gradients.
+        n_interp_half = 75
```
albangossard, line 85: For the moment we can keep this parameter hard-coded, but in the future this is definitely an hyper-parameter we want to play with.

One question I would like to bring here is where do we want these parameters to be stored? If it only impacts the gradient computation, I think we should introduce a new VJP flavor and store that information in there. Otherwise, if we also want to use interpolation in the forward, this should be inside UDE parameters in my opinion. What are your thoughts about this?

src/models/target/target_D.jl
```diff
@@ -0,0 +1,229 @@
...
+        # We construct an interpolator with quantiles and equal-spaced points
+        H_interp_unif = LinRange(0.0, maximum(H), n_interp_half) |> collect
+        H_interp_quantiles = quantile!(H[H .> 0.0], LinRange(0.0, 1.0, n_interp_half))
+        H_interp = vcat(H_interp_unif, H_interp_quantiles)
+        H_interp = unique(H_interp)
+        H_interp = sort(H_interp)
```
albangossard, line 91: I really like this implementation, that's neat!

src/models/target/target_D.jl
```diff
@@ -0,0 +1,229 @@
...
+
+        # Compute exact gradient in certain values of H
+        grads = []
+        # TODO: Check if all these gradints cannot be computed at once withing Lux
```
albangossard, line 95: I think it's doable by batching, isn't it?

src/models/target/target_D.jl
```diff
@@ -0,0 +1,229 @@
...
+
+function ∂D∂H_target_D(; H, ∇S, θ, ice_model, ml_model, glacier, params)
+    n = ice_model.n
+    Γ_no_A = Γ(ice_model, params; include_A = false)
+    A = _apply_parametrization_A_target_D(; H, ∇S, θ, ice_model, ml_model, glacier, params)
+    ∂D∂H_no_NN = (n[] + 2) .* A .* Γ_no_A .* H.^(n[] + 1) .* ∇S.^(n[] - 1)
+
+    # Derivative of the output of the NN with respect to input layer
+    # TODO: Change this to be done with AD or have this as an extra parameter.
+    # This is done already in SphereUDE.jl with Lux
```
albangossard, line 39: Yes I think this is something that is worth implementing, just to avoid having to play with an extra epsilon parameter to compute the gradient :)
No need to implement it right now, I created a task for that so that we keep in mind we need to implement this https://github.com/ODINN-SciML/ODINN.jl/issues/306
We can move forward with this PR and replace with AD in main later on.

src/models/target/target_D.jl
```diff
@@ -0,0 +1,229 @@
+### Target to invert D as a function of H and Temp
+
+"""
+    build_target_D()
+
+Inversion of the form
+
+    D(H, ∇S, θ) = 2 / (n + 2) * (ρg)^n H^{n+2} |∇S|^{n-1} * NeuralNet(T, H; θ)
+    D(H, ∇S, θ) = 2 / (n + 2) * (ρg)^n H^{n+2} |∇S|^{n-1} * exp( NeuralNet(T, H; θ))
```
albangossard, line 9: I guess this is the first one that you've implemented, right?

src/models/machine_learning/MLmodel.jl
```diff
@@ -65,20 +66,45 @@ Creates a new feed-forward neural network.
...
+        @warn "To specify the neural network please provide all (architecture, θ, st), not just a subset of them."
     end
 
+    # Build target based on parameters
+    target_object = SIA2D_target(name = params.UDE.target)
+
     # Build the simulation parameters based on input values
-    neural_net = NeuralNetwork{typeof(architecture), typeof(θ), typeof(st)}(architecture, θ, st)
+    # TODO: I don't think target should be inside NN, but rather having the neural net elements
+    # as part of a regressor type.
```
albangossard, line 103: At first I agreed with your comment but the more I think about it, the more I realize it's not really blocking.
We plan to implement the `Law` interface where the user only provides a regressor and eventually the weights (see `f` in https://github.com/ODINN-SciML/ODINN.jl/issues/302). This corresponds to providing only `architecture` and `theta`. We might have to think on our side about how we plan to expose `st` to the user but the real problem is how do we handle different types of regressor? Here it's a neural network but we could use a Gaussian Process or any other kind of regressor, and the splitting between `architecture, `theta` and `st` is really specific to the NN.

src/inverse/SIA2D/gradient.jl
```diff
@@ -196,6 +201,8 @@ end
 
 """
 Define SIA2D forward map for the adjoint mode
+
+This is currently just use for Enzyme
```
albangossard, line 205: In the end do we really need this or could it be mutualized? This is something @JordiBolibar mentioned yesterday
By the way, the function `apply_UDE_parametrization_enzyme` isn't used anywhere and it can be removed.

src/simulations/functional_inversions/functional_inversion_utils.jl
```diff
@@ -106,15 +105,17 @@ function train_UDE!(simulation::FunctionalInversion, optimizer::Optim.FirstOrder
...
-        cb_plots = (θ, l) -> false
-    elseif simulation.parameters.UDE.target == "A"
-        cb_plots = (θ, l) -> false
-        # This other option returns weird error right now, commenting for now
-        # cb_plots(θ, l) = callback_plots_A(θ, l, simulation) # TODO: make this more customizable
-    else
-        raise("Simulation target not defined.")
-    end
+    cb_plots = (θ, l) -> false
+    # TODO: Do we need this following code?
```
albangossard, line 109: I don't think so, generating plots during run time is a bad practice, so I'm in favor of removing it.

src/simulations/functional_inversions/functional_inversion_utils.jl
```diff
@@ -229,63 +233,65 @@ function loss_iceflow_transient(θ, simulation::FunctionalInversion)
...
-    elseif loss_type == "V"
-        l_tot = l_V/length(simulation.results)
-    elseif loss_type == "HV"
-        l_tot = (l_V + l_H)/length(simulation.results)
-    end
-
-    return l_tot
-    end # let
-end
+# TODO: This function is not longer use in the new version of the code with transient
```
albangossard, line 236: Yes I'm looking forward to implementing this with the new ice velocity structs!

Maybe we can comment this piece of code for the moment? Just to be sure to implement something that is mathematically equivalent to what you did in the first ODINN paper.

src/simulations/functional_inversions/functional_inversion_utils.jl
```diff
@@ -361,10 +368,7 @@ function simulate_iceflow_UDE!(
     model = simulation.model
     params = simulation.parameters
 
-    # TODO: make this more general
-    # @infiltrate
-    apply_UDE_parametrization!(θ, simulation, nothing, batch_id) # Apply the parametrization otherwise the physical values are wrong between the beginning of the simulation and the first callback
```
albangossard, line 366: In the end don't we need this for MB since it's a discrete callback?

albangossard (commented):
Thanks a LOT @facusapienza21 for this huge piece of work! I'm happy to see this interface taking form. Please find some comments below.

Also I haven't checked the maths of all the backward functions, but if you think this deserves a double check, I can do it.

src/models/machine_learning/MLmodel.jl
JordiBolibar: As discussed yesterday, the `MLmodel` which goes inside `Model` would go, and `Target` would be placed inside `Law` or merged with it.

src/models/machine_learning/MLmodel.jl
JordiBolibar: Yes, all this will require some thinking. Nonetheless, for the sake of simplicity and to focus on the scientific implementation, I think that the high-level API should be flexible enough to accommodate this with minor changes. Therefore, I suggest we move one with what we discussed yesterday, focussing on `Chain`, and then when we want to make it more flexible we think about which extra changes need to be done.

src/simulations/functional_inversions/functional_inversion_utils.jl
JordiBolibar: I think this is to plot the evolution of the training. I'm not against it, as long as the plot is not displayed and is just stored. @albangossard do you think it would be better to store everything in memory and then make all the plots at the end? For that we would have to change a bit the implemented logic. But to be honest, we haven't used those plot callbacks in such a long time that I'm sure they must be broken.

albangossard: Yes in terms of performance it's much better to store everything in memory and make the plot in the post-processing. I created a dedicated issue for this https://github.com/ODINN-SciML/ODINN.jl/issues/307, so that we don't have to worry about it for now.

src/simulations/functional_inversions/functional_inversion_utils.jl
albangossard: We used to call this `apply_UDE_parametrization!` function here because between the beginning of the simulation and the first time of `tstops`, the mass balance wasn't initialized. This is specific to the terms that are initialized with a `DiscreteCallback`.
From what I understand from the implementation in this PR, this is correctly handled for the other variabels because we call the different neural network components within SIA2D. But as long as we use `DiscreteCallback`, we still need to call the mass balance neural network for `t=tspan[1]` or am I wrong?

albangossard: Sorry I realize that my sentence was very confusing. Yes I mean the neural network that used to be synced with the same callback as the one used for MB.

test/runtests.jl
```diff
@@ -37,32 +40,36 @@ ENV["GKSwstype"]="nul"
...
-
-@testset "Manual implementation of the continuous adjoint with Enzyme VJP vs finite differences" test_grad_finite_diff(ContinuousAdjoint(VJP_method=ODINN.EnzymeVJP()); thres=[2e-4, 1e-8, 1e-3])
-
-@testset "Manual backward of the loss terms vs Enzyme" test_grad_loss_term()
+@testset "Adjoint of unit operations inside SIA2D" begin
+    @testset "Adjoint of diff" test_adjoint_diff()
+    @testset "Adjoint of clamp_borders" test_adjoint_clamp_borders()
+    @testset "Adjoint of avg" test_adjoint_avg()
+    # I don't think we need the following test anymore since it is done later
+    # @testset "Adjoint of SIA2D" test_adjoint_SIAD2D()
```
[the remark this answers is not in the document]
facusapienza21: Sorry, I didn't explain well what I meant in the comment. I saw the code for `test_adjoint_SIA2D()` and it seems to be running the same tests that are being run later. I guess a better question is: what is this function testing that is not being tested in `test_grad_loss.jl`? I agree is good to have unit tests at different levels, but this one seems to close to the others. Am I missing something? Happy to fix these tests, I just had the idea they are "repeated".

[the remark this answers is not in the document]
albangossard: Do you mean that `test_adjoint_SIAD2D()` checks the same piece of code as `test_grad_finite_diff()`? For me these two tests are very different since the first one tests only the VJP while the second one also tests the integration over time. If the first one passes but the second one fails, you know that the problem is in the integration with respect to time. On the other hand, if both fail, it's likely because the VJP is broken. If we break the gradient computation, this is very useful to have these two tests.
Did I understand well what you meant or am I missing something and the tests could be speeded-up?

src/models/target/target_D.jl
facusapienza21: Yes! Done.

src/models/target/target_D.jl
facusapienza21: Yes! I just did.

src/models/target/target_D.jl
facusapienza21: This is a good point. I am not sure what is the best place for this, since this is also very dependent of the Target itself. Just to make a little step forward in addressing this (but not fully solving it), I moved this hardcoded parameter to the function definition and to the `Target` constructor before that, so these parameters are more explicit and easy to "see" and handle for the higher lever API.

src/models/target/target_D.jl
facusapienza21: What do you mean here @albangossard ?

src/models/target/target_D.jl
facusapienza21: Yes, the other one was just a side note for myself. I remove the comment from the code.

src/inverse/SIA2D/gradient.jl
facusapienza21: Nice! I get rid of the `apply_UDE_parametrization_enzyme` (no bang).

src/simulations/functional_inversions/functional_inversion_utils.jl
facusapienza21: Sounds good to me: I also never really liked the callback to make plots. Should we still keep the code for the set callback 
```
cb_plots = (θ, l) -> false
cb_diagnosis(θ, l) = callback_diagnosis(θ, l, only(simulation_train_loader.data))
cb(θ, l) = CallbackOptimizationSet(θ, l; callbacks=(cb_plots, cb_diagnosis))
```
to have some freedom or better just to do `cb` to simply be ``cb_diagnosis`? I am fine any ways.

src/simulations/functional_inversions/functional_inversion_utils.jl
facusapienza21: Yes, I agree. We should also move all the loss functions outside `functional_inversion_utils.jl`, don't you think?

facusapienza21:
@JordiBolibar would you please take a quick look at then Enzyme code? I didn't make major changes to the existing code with Enzyme, but this still breaks. Happy to debug this more, but since you spend more time with this you can give me a clue. 

Right now the VJPs with Enzyme are computed as 
```julia
function VJP_λ_∂SIA∂H(VJPMode::EnzymeVJP, λ, H, θ, simulation, t, batch_id)
    dH_H = Enzyme.make_zero(H)
    λ_∂f∂H = Enzyme.make_zero(H)
    _simulation = Enzyme.make_zero(simulation)
    smodel = StatefulLuxLayer{true}(simulation.model.machine_learning.architecture, θ.θ, simulation.model.machine_learning.st)

    λH = deepcopy(λ) # Need to copy because Enzyme changes the backward gradient in-place
    Enzyme.autodiff(
        Reverse, SIA2D_UDE!, Const,
        # Reverse, SIA2D_adjoint!, Const,
        Enzyme.Const(θ),
        Duplicated(dH_H, λH),
        Duplicated(H, λ_∂f∂H),
        Enzyme.Duplicated(simulation, _simulation),
        Enzyme.Const(smodel),
        Enzyme.Const(t),
        Enzyme.Const(batch_id)
    )
    return λ_∂f∂H, dH_H
end
```
with ` SIA2D_UDE!` being the function 
```julia
function SIA2D_UDE!(_θ, _dH::Matrix{R}, _H::Matrix{R}, simulation::FunctionalInversion, smodel, t::R, batch_id::I) where {R <: Real, I <: Integer}

    # We load the ML model with the parameters
    smodel.ps = _θ.θ
    smodel.st = simulation.model.machine_learning.st

    simulation.model.machine_learning.target.apply_parametrization!(;
        H = _H, ∇S = nothing, θ = _θ,
        ice_model = simulation.model.iceflow[batch_id],
        ml_model = simulation.model.machine_learning,
        glacier = simulation.glaciers[batch_id],
        params = simulation.parameters
    )

    Huginn.SIA2D!(_dH, _H, simulation, t; batch_id = batch_id)

    return nothing
end
```
I am here trying to use the new API for apply parametrization, so this will depend of the target objective. For the inversions of A, this function looks like this: 
```julia
function apply_parametrization_target_A(; H, ∇S, θ, ice_model, ml_model, glacier, params)
# function apply_parametrization_target_A(H, ∇S, θ, ice_model, ml_model, params, glacier) where {I <: Integer, SIM <: Simulation}
    # We load the ML model with the parameters
    nn_model = ml_model.architecture
    st = ml_model.st
    smodel = StatefulLuxLayer{true}(nn_model, θ.θ, st)
    min_NN = params.physical.minA
    max_NN = params.physical.maxA
    A = predict_A̅(smodel, [mean(glacier.climate.longterm_temps)], (min_NN, max_NN))[1]
    return A
end

function apply_parametrization_target_A!(; H, ∇S, θ, ice_model, ml_model, glacier, params)
    A = apply_parametrization_target_A(; H, ∇S, θ, ice_model, ml_model, glacier, params)
    ice_model.A[] = A
    ice_model.D = nothing
    ice_model.D_is_provided = false
    return nothing
end
```
From my perspective, what was Const and Duplicated hasn't change from the previous version of the code to the new one, reason why I am confused with the error message that Enzyme gives me: 
```
ERROR: LoadError: Constant memory is stored (or returned) to a differentiable variable.
As a result, Enzyme cannot provably ensure correctness and throws this error.
This might be due to the use of a constant variable as temporary storage for active memory (https://enzyme.mit.edu/julia/stable/faq/#Runtime-Activity).
If Enzyme should be able to prove this use non-differentable, open an issue!
To work around this issue, either:
 a) rewrite this variable to not be conditionally active (fastest, but requires a code change), or
 b) set the Enzyme mode to turn on runtime activity (e.g. autodiff(set_runtime_activity(Reverse), ...) ). This will maintain correctness, but may slightly reduce performance.
Mismatched activity for:   store {} addrspace(10)* %.fca.0.extract, {} addrspace(10)* addrspace(10)* %42, align 8, !dbg !152, !tbaa !182, !alias.scope !135, !noalias !136 const val:   %.fca.0.extract = extractvalue { {} addrspace(10)* } %0, 0, !dbg !66, !enzyme_type !67, !enzymejl_byref_MUT_REF !0, !enzymejl_source_type_Vector\7BFloat64\7D !0
 value=Unknown object of type Vector{Float64}
 llvalue=  %.fca.0.extract = extractvalue { {} addrspace(10)* } %0, 0, !dbg !66, !enzyme_type !67, !enzymejl_byref_MUT_REF !0, !enzymejl_source_type_Vector\7BFloat64\7D !0
```

Let me see if you see something that looks obviously wrong and I can try to fix it (end of my day here and maybe you have a clear direction on how to fix this!)

JordiBolibar:
@facusapienza21 OK, let me take a look. Yesterday I started debugging the Enzyme code again since it bothers me so much to abandon that path. I will move to your branch and start again.

src/models/target/target_D.jl
albangossard: Instead of doing a loop, I was thinking of concatenating all the $(H[i,j])_{i,j}$ in a vectorial form $(H[k])_k$ and treat them as batches to backpropagate and get the gradient of theta for each element of the batch. I see that a very smart guy already wrote [some part of the documentation](https://lux.csail.mit.edu/stable/manual/nested_autodiff#Using-Batched-Jacobian-for-Multiple-Inputs) about `batched_jacobian`. I guess this function will also work for batched parameters instead of batched inputs since Zygote should not see the difference between these two.

src/simulations/functional_inversions/functional_inversion_utils.jl
albangossard: Yes I agree they should be in a dedicated file in `ODINN/src/losses/`

src/simulations/functional_inversions/functional_inversion_utils.jl
albangossard: I think that in the future we can log everything in `cb_diagnosis`, so we probably don't need a `CallbackOptimizationSet`.

facusapienza21: Sounds good! Just get rid of it.

facusapienza21:
I changed the API to be now based on multiple distpatch. The target object and all the code looks much clean now and I imagine this will serve better in order to merge with the higher API idea. 

Something I couldn't make work and maybe @albangossard @JordiBolibar can help me is with the multiprocessing. I notice that Sleipnir now has defaults in the parameters constructors `workers::I = 4`, which breaks my tests in local. All starts working when I set `workers::I = 1` again. I didn't have a good change to take a look at this, but this is why the tests are now breaking here. Let me know if you have an idea of how to fix this.

albangossard (commented):
@facusapienza21 I tried to run the tests with your latest changes and on my side the tests start failing at `test_adjoint_SIAD2D()`. It seems that with the current design, it's not possible to compute the VJP without having to define a ML model. In my opinion this is a bit too restrictive as it does not allow us to run simple inversions (not a functional inversion). Do you have an idea of how we could avoid having to define a ML model with your API?

facusapienza21:
Hi @albangossard ! Thank you for the comment. I fixed this in the last commit, where I restored the VJPs test to work again. A model of \theta needs to be provided in order to compute \partial \theta, but I agree that this is not required for dH. Now the tests 
```julia
@testset "Discrete VJP of SIA2D vs finite differences" test_adjoint_SIA2D(ContinuousAdjoint(VJP_method = DiscreteVJP()); target = :A)
@testset "Continuous VJP of SIA2D vs finite differences" test_adjoint_SIA2D(ContinuousAdjoint(VJP_method = ContinuousVJP()); target = :A)
```
should pass. I am still confused why the test ` test_adjoint_SIA2D()` is repeated in the tests under "Adjoint of unit operations inside SIA2D" and "Adjoint method of SIA equation with A as target". Would it have more sense to include them all in the first category (plus the Enzyme VJP)?

src/models/machine_learning/MLmodel.jl
```diff
@@ -65,20 +66,61 @@ Creates a new feed-forward neural network.
...
+            architecture = build_default_NN(; n_input = 2, lightNN = lightNN)
+            architecture, θ, st = set_NN(architecture; θ_trained = θ, ft = ft)
+        elseif params.UDE.target == :D
+            @warn "Inversion for :D has not been setup yet."
+            architecture = build_default_NN(; n_input = 2, lightNN = lightNN)
+            architecture, θ, st = set_NN(architecture; θ_trained = θ, ft = ft)
+        else
+            @warn "Constructing default Neural Network"
+            architecture, θ, st = get_default_NN(θ, ft; lightNN = lightNN)
+        end
```
albangossard, line 101: It could be simplified by adding `n_input` as an argument to `get_default_NN` and then simply setting the value of `n_input` based on `params.UDE.target`. I know that this makes the NN returned by `get_default_NN` "non-unique" but I don't think it really breaks the logic as the structure of the NN remains the same after the first layer.

test/SIA2D_adjoint_utils.jl
```diff
@@ -0,0 +1,250 @@
...
+#     thres_relerr = 1e-14
+#     if printDebug | !( (min_ratio<thres_ratio) & (min_angle<thres_angle) & (min_relerr<thres_relerr) )
+#         println("Gradient wrt θ")
+#         println("eps    = ",printVecScientific(eps))
+#         printVecScientific("ratio  = ",ratio,thres_ratio)
+#         printVecScientific("angle  = ",angle,thres_angle)
+#         printVecScientific("relerr = ",relerr,thres_relerr)
+#     end
+#     @test (min_ratio<thres_ratio) & (min_angle<thres_angle) & (min_relerr<thres_relerr)
+# end
```
albangossard, line 250: This is already in the git history of Huginn, so I think we can remove it entirely :)

test/runtests.jl
```diff
@@ -21,48 +21,56 @@ using SciMLSensitivity
...
-@testset "Backward of the loss terms vs Enzyme" test_grad_loss_term()
+@testset "Adjoint of unit operations inside SIA2D" begin
+    @testset "Adjoint of diff" test_adjoint_diff()
+    @testset "Adjoint of clamp_borders" test_adjoint_clamp_borders()
+    @testset "Adjoint of avg" test_adjoint_avg()
+    # @testset "Adjoint of SIA2D" test_adjoint_SIAD2D()
+end
 
-@testset "Implementation of continuous adjoint with discrete VJP vs Enzyme for Halfar solution" test_grad_Halfar(ContinuousAdjoint(VJP_method=DiscreteVJP()); thres=[3e-1, 1e-15, 5e-1])
+@testset "Adjoint method of SIA equation with A as target" begin
```
albangossard, line 56: I didn't know that nested testsets automatically adjust their prints depending if a dependency fails or not. I like how it renders in the REPL!

albangossard (approved):
Thanks a lot @facusapienza21 for this huge work! I can't wait to play with the inversions!

Please find some minor comments below, mainly to improve code readability.
There seems to be an issue with the gradient computation for all the `test_grad_finite_diff` tests. There might be a bug that affects all these tests at the same time.
That's a little sad but I must admit that we won't be able to fix the Enzyme tests before we merge your PR, so let's just comment them.

Once the `test_grad_finite_diff` tests are fixed, this can be merged. Very good job!

src/inverse/SIA2D/adjoint.jl
```diff
@@ -0,0 +1,407 @@
...
+    # Compute diffusivity based on target objective
+    # D = Diffusivity(
+    #     target;
+    #     H = H̄, ∇S = ∇S, θ = θ,
+    #     iceflow_model = SIA2D_model, ml_model = ml_model,
+    #     glacier = glacier, params = params
+    # )
+
+    # Gradient wrt θ
+    ∂D∂θ = ∂Diffusivity∂θ(
```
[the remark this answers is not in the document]
facusapienza21: Ja this was @JordiBolibar ideas... I am a bad programmer.

facusapienza21:
@albangossard I change the inversion tests to see if now they pass. Also notice I added new tests for the pure D inversion, so now we have three different inversions working: A, a more hybrid A-D like, and a pure D inversion. The first two have been tested and I am sure they work. The last one I finished today and it seems to pass the gradient tests but I still haven't run a long simulation to see if it learns D as it should. 

Let me know what you think. If you approve, I will merge after (don't merge yet, since I want to take a look at a simulation first, but let me know if you approve the merge).

Also, what do you think about bumping the version of ODINN after this (third digit). It seems we made a lot of changes here and the simulations are running, so it will nice to be able to install and run. I can also make some cleaning on the `scripts` and `examples` before merging if that is desirable.

albangossard (commented):
Good idea to add more tests! This covers well all the use-cases.
I think that soon we will have to consider designing a toy glacier with very few data points to speed-up the tests. A CI that runs in 3.5h is too much and it doesn't allow us to develop and patch things quickly. With Gungnir we could generate a glacier that have a grid of size 10x12 and run the tests on that glacier. What do you think? No need to implement that in this PR, it's just to discuss how we can speed-up the tests.

I have minor concerns, otherwise this looks good to me. After my comments are addressed, this PR can be merged.

I know all of this takes time but I would like to merge https://github.com/ODINN-SciML/Huginn.jl/pull/83 on Friday at latest. That would be great if you can merge your PR before, otherwise I'll merge mine anyway as having to handle the merge conflicts later on will be annoying.

I'm ok to bump ODINN's version! :rocket:

facusapienza21:
Hi @albangossard ! Thanks for the comments, I will address them all soon and merge this PR. 

I completely agree with your comment on having a small glacier (I actually thought about this yesterdays and I didn't do it, but definitively I agree on doing this since testing takes a lot!).

src/models/target/target_D_pure.jl
```diff
@@ -0,0 +1,308 @@
...
+    return ∂D∂H_no_NN + ∂D∂H_NN
+end
+
+function ∂Diffusivity∂∇H(
+    target::SIA2D_D_target;
+    H, ∇S, θ, iceflow_model, ml_model, glacier, params
+    )
+
+    # For now we ignore the derivative in surface slope
+    δ∇H = 1e-6 .* ones(size(∇S))
```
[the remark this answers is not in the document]
albangossard: Ok thank you for this explanation!

facusapienza21:
@albangossard I fixed the original problem with the docs (which actually forced me to change the API, which I don't know if it is a desirable behaviour of the docs) but now the inversion test of the docs does not run (since I change a little this API). I will make this change tomorrow and since I already addressed your comments in this PR, once the docs runs I will merge and bump the version!

PR #304, opened 2025-04-18
Functional inversion tutorial

albangossard:
Follows https://github.com/ODINN-SciML/ODINN.jl/pull/299 with a fix of the documentation generation.


docs/make.jl
```diff
@@ -27,7 +27,7 @@ tutorial_files = [
 # Generate independent Markdown files for each tutorial
 for tutorial_file in tutorial_files
     tutorial_name = splitext(basename(tutorial_file))[1]  # Extract the file name without extension
-    Literate.markdown(tutorial_file, "./src"; name = tutorial_name)
+    Literate.markdown(tutorial_file, "./src"; name = tutorial_name, preprocess = replace_includes)
```
JordiBolibar, line 30: Is this still needed? I had forgotten about this util, so I moved everything to `src`, and I didn't seem to have issues then. If we keep this function then we might as well move everything back to the `tutorials` folder. 

Maybe this is what was needed to fix it? Good thing you took a look, I had completely forgotten about this haha.

albangossard: I first reverted many changes to make the documentation work but it turns out that `replace_includes` is no longer needed. I'll remove that function in an upcoming commit.

For the moment `makedocs` freezes when we use multiprocessing.

albangossard:
In the end it wasn't possible to make the documentation generation work. This problem seems specific to Ubuntu and the CI. What we did is that we added an environment variable `ODINN_OVERWRITE_MULTI` that overwrites the behavior given by the simulation parameters in Sleipnir (see https://github.com/ODINN-SciML/Sleipnir.jl/pull/102). This allows the tutorial to apparently work with the multiprocessing as we're providing `multiprocessing=true` but under the hood, multiprocessing is disabled.

@facusapienza21 I made releases of the different packages in order to be able to build the documentation in this PR.

PR #310, opened 2025-04-24
Make SIA2D_target type stable

JordiBolibar:
Changes to make the functions in SIA2D_target parametric, as well as a new script to start debugging Enzyme


facusapienza21 (approved):
I just change a few things in the syntax and make sure that the test are working. I will proceed to merge this with the other branch and see if I can fix the Enzyme issues there.

PR #317, opened 2025-05-12
[WIP] Design of pipeline to save results from ML inversion

facusapienza21:
@albangossard as I am running simulation for the inversion of the diffusivity D, I thought it was going to be useful to figure out how to save relevant outputs of the simulation so these can be analyze later. The most important of these, the trained parameters of the neural network to allow post evaluation. This PR is trying to address this.

Notice that Sleipnir also has a save results function, but this is rather different in nature than saving the information of the model (rather than the information of a forward SIA simulation). What do you think? 

This PR is work in progress, so it would be nice to think (assuming we approved the API design) what else we would like to save. Notice that the architecture of the NN will be harder to save, so maybe we want to think about this. 

So far, this PR has

- Design of simple Results object
- Save of such object at the end of simulation in `jld2` format
- A test to assess that the file is saving the information correctly

Any feedback welcome!

### DIFF
--- a/src/results/Results.jl
+++ b/src/results/Results.jl
@@ -1,0 +1,12 @@
+export Result
+export save_simulation_file!
+
+abstract type AbstractResult end
+
+@kwdef struct Result{F <: AbstractFloat} <: AbstractResult
+    θ::ComponentVector
+    losses::Vector{F}
+    params::Sleipnir.Parameters
+end
+
+include("result_utils.jl")
--- a/src/results/result_utils.jl
+++ b/src/results/result_utils.jl
@@ -1,0 +1,32 @@
+"""
+
+This function saves the results of a simulation to a file in JLD2 format. If the `path` argument is not provided, the function will create a default path based on the current project directory. The results are saved in a file named `prediction_<nglaciers>glaciers_<tspan>.jld2`, where `<nglaciers>` is the number of glaciers in the simulation and `<tspan>` is the simulation time span.
+"""
+function save_simulation_file!(
+    sol,
+    simulation::SIM;
+    path::Union{String,Nothing} = nothing,
+    file_name::Union{String,Nothing} = nothing
+    ) where {SIM <: Simulation}
+
+    # Create path for simulation results
+    if isnothing(path)
+        simulation_path = joinpath(dirname(Base.current_project()), "data/results/simulation")
+    else
+        simulation_path = path
+    end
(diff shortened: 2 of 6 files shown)


src/results/result_utils.jl
```diff
@@ -0,0 +1,32 @@
...
+function save_simulation_file!(
+    sol,
+    simulation::SIM;
+    path::Union{String,Nothing} = nothing,
+    file_name::Union{String,Nothing} = nothing
+    ) where {SIM <: Simulation}
+
+    # Create path for simulation results
+    if isnothing(path)
+        simulation_path = joinpath(dirname(Base.current_project()), "data/results/simulation")
```
albangossard, line 14: Maybe something more precise than "simulation" which could be either an inversion or a prediction? Currently, ODINN already generates a "data/results/predictions" folder

src/results/Results.jl
```diff
@@ -0,0 +1,12 @@
+export Result
+export save_simulation_file!
+
+abstract type AbstractResult end
+
+@kwdef struct Result{F <: AbstractFloat} <: AbstractResult
+    θ::ComponentVector
+    losses::Vector{F}
+    params::Sleipnir.Parameters
```
albangossard, line 9: We could add some stats about the performance like the time per epoch, what do you think?

I'm even wondering if we should rely on an external library that does everything for us. Tensorboard is the gold standard to log learning histories in Python and there is a Julia backend: https://github.com/JuliaLogging/TensorBoardLogger.jl
That would allow us to access many stats like the time per step, and the training/validation loss seamlessly. What do you think?

albangossard (commented):
Thank you @facusapienza21, that's a good start to start logging information about the training!
I'm wondering if we should use https://github.com/JuliaLogging/TensorBoardLogger.jl to log information like the training histories. That would make our life easier.
That's not incompatible with the `Results` struct that you created, it's just complementary. We could use tensorboard to log the stats and loss histories, and use `Results` to collect the weights and the parameters in order to run predictions afterwards after the training.
What is your opinion on this?

facusapienza21:
Last commit closes #339 and restores the inversion test. Test with MB and multiprocessing should also being included in the future.

JordiBolibar:
Hi, I'm joining the party quite late, but why name this new struct `Results` if we already have one called `Results` which does something different? If I understood correctly one is for the simulation results, and the other one is for the model itself. Maybe we should rename it?

PR #341, opened 2025-06-05
Ice velocity loss

albangossard:


src/losses/Losses.jl
```diff
@@ -1,54 +1,404 @@
...
-    return [backward_loss(lossType, ai, bi; normalization=normalization) for (ai,bi) in zip(a,b)]
+function backward_loss(
+    lossType::LogSum,
+    a::Matrix{F},
+    b::Matrix{F},
+    mask::BitMatrix;
+    normalization::F,
+) where {F <: AbstractFloat}
+    d = zero(a)
+    d[mask] = log.((a[mask] .+ lossType.ϵ) ./ (b[mask] .+ lossType.ϵ)) ./ (b[mask] .+ lossType.ϵ)
```
albangossard, line 210: Have you tested this backward? Assuming that the function is being differentiated wrt `a`, when I write down the equations I have a different formula.
```suggestion
    d[mask] = log.((a[mask] .+ lossType.ϵ) ./ (b[mask] .+ lossType.ϵ)) ./ (a[mask] .+ lossType.ϵ)
```

src/losses/Losses.jl
```diff
@@ -1,54 +1,404 @@
...
+    @assert (minimum(a) >= 0.0) & (minimum(b) >= 0.0)
+    return sum(log.((a .+ lossType.ϵ) ./ (b .+ lossType.ϵ)).^2)
+end
+function backward_loss(
+    lossType::LogSum,
+    a::Matrix{F},
+    b::Matrix{F};
+    normalization::F,
+) where {F <: AbstractFloat}
+    return 2.0 .* log.((a .+ lossType.ϵ) ./ (b .+ lossType.ϵ)) ./ (b .+ lossType.ϵ)
```
albangossard, line 229: Same thing here

src/losses/Losses.jl
```diff
@@ -1,54 +1,404 @@
...
+end
+
+function loss(
+    lossType::LogSum,
+    a::Matrix{F},
+    b::Matrix{F};
+    normalization::F=1.,
+) where {F <: AbstractFloat}
+    @assert (minimum(a) >= 0.0) & (minimum(b) >= 0.0)
+    return sum(log.((a .+ lossType.ϵ) ./ (b .+ lossType.ϵ)).^2)
```
albangossard, line 221: Normalization is missing, same thing in the backward

src/losses/Losses.jl
albangossard: By the way we should test this new loss function by adding a test in the `"Manual backward of the loss terms vs Enzyme"` testset

facusapienza21: This is embarrassing @albangossard ... all this automatic differentiation and I made a simple manual differentiation error. I am a shame, you should stop collaborating with me... I fix this now and tests are passing XD

PR #347, opened 2025-07-03
Update for laws and plotting utils

JordiBolibar:
This PR is ongoing work to start adding plotting utils for laws, and also to make some small changes to include other behaviours, e.g. easily switching from scalar to matrix outputs. 

Feel free to comment or suggest/make commits!

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -20,6 +20,8 @@ using Documenter, Literate
 using ODINN
 using DocumenterCitations
 
+cd(dirname(Base.active_project()))
+
 bib = CitationBibliography(
     joinpath(@__DIR__, "src/assets", "references.bib");
     style=:numeric
--- a/docs/src/laws.jl
+++ b/docs/src/laws.jl
@@ -63,9 +63,7 @@ rgi_paths = get_rgi_paths()
 rgi_ids = ["RGI60-11.03638"]
 δt = 1/12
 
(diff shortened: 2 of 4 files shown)


src/laws/laws_utils.jl
```diff
@@ -24,16 +24,18 @@ Evaluates a law on the specified glacier within a simulation context and for a u
...
+    @show scalar
+    cache = init_cache(law, simulation, glacier_idx, params; scalar=scalar)
     if !isnothing(simulation.model.machine_learning)
         simulation.model.machine_learning.θ = θ
     end
 
-    law.f.f(cache, inputs, θ)
+    law.f.f(cache, input_values, θ)
     return cache
 end
```
mleseach, line 40: Since this function is not performance-critical, I think we can simply create a zero-dimensional `cache` manually and get rid of `simulation`, `glacier_idx`, and `scalar`.

```julia
function eval_law(law::AbstractLaw, input_values::NamedTuple, θ)
    # Initialize the cache to be able to perform inference on the law
    cache = zeros(eltype(cache_type(law)))

    law.f.f(cache, input_values, θ)
    return cache[]
end
```

What do you think?

JordiBolibar:
> @JordiBolibar I let you handle the rebase/merge before I review this because currently it's hard to read the history

It seems OK on my side? Can you check again? The last commit included the rebase.

albangossard (commented):
> It seems OK on my side? Can you check again?

Yes it works, sorry when I tried to review the target branch was `main` and not `dev`. That's why there used to be many changes.

PR #350, opened 2025-07-07
Fix Enzyme VJPs + make compatible with SciML

albangossard:
This PR includes the changes of https://github.com/ODINN-SciML/ODINN.jl/pull/346 + some updates to make our code compatible with SciML


src/simulations/functional_inversions/functional_inversion_utils.jl
```diff
@@ -266,132 +293,232 @@ end
...
-        θ,
-        simulation::SIM,
+        container::FunctionalInversionBinder,
         cb::SciMLBase.DECallback,
-        glacier_idx::I,
-    ) where {I <: Integer, SIM <: Simulation}
+        iceflow_prob::ODEProblem,
+    )
 
 Make forward simulation of the iceflow UDE.
```
JordiBolibar, line 464: This should be "Make a forward simulation..."

src/simulations/functional_inversions/functional_inversion_utils.jl
```diff
@@ -266,132 +293,232 @@ end
...
-        cache.iceflow.H₀,
-        params.simulation.tspan,
-        θ;
-        tstops=params.solver.tstops,
-    )
-
+    iceflow_prob::ODEProblem,
+)
+    params = container.simulation.parameters
+    iceflow_prob_remake = remake(iceflow_prob; p=container)
```
JordiBolibar, line 472: So just to follow, one of the main changes to adapt this is to avoid using a closure and instead use a `remake` to pass the SciMLStruct with the simulation and the parameters?

JordiBolibar (commented):
Mostly comments to better understand the implementation of the SciML fix. It would be nice to properly document some of these things and choices. I know you have added already some stuff in the differentiability, but probably there will be room to add a bit more details.

PR #355, opened 2025-07-23
Enable SciMLSensitivity gradient computation test

albangossard:
Enable the SciMLSensitivity gradient computation test. I temporarily added lines to use in dev mode [my fork](https://github.com/albangossard/SciMLSensitivity.jl) that implements the fix for scimlstructs.
**We should not** make a release of ODINN without removing these lines.


facusapienza21:
This looks like a good fix to me @albangossard ! I think we can merge this if you considere it appropriate!

PR #358, opened 2025-08-06
Ice velocity loss merge conflict

albangossard:


test/test_grad_loss.jl
```diff
@@ -135,7 +135,20 @@ function test_grad_finite_diff(
...
+    try
+        loss_iceflow_grad!(dθ, θ, simulation)
+    catch
+        try
+            loss_iceflow_grad!(dθ, θ, simulation)
+        catch
+            loss_iceflow_grad!(dθ, θ, simulation)
+        end
+    end
+    ######
```
facusapienza21, line 151: I think is good to have this, but I will add a big warning if this executes under the `catch` statement just to provide some information about the error

PR #359, opened 2025-08-07
Introducing Regularization into Training

facusapienza21:


src/losses/Regularization.jl
```diff
@@ -0,0 +1,155 @@
...
+
+    # TODO: extract this from glacier
+    Δx, Δy = 1.0, 1.0
+
+    if !isnothing(simulation.model.machine_learning)
+        simulation.model.machine_learning.θ = θ
+    end
+    Vx, Vy, V = Huginn.V_from_H(simulation, H, t, θ)
+
+    mask = is_in_glacier(H, 3) .& (V .> 0.0)
```
albangossard, line 90: I know this is still a draft, in the future we should make `distance` a field of the regularization type

src/models/target/target_D_hybrid.jl
```diff
@@ -117,6 +116,7 @@ function ∂Diffusivity∂θ(
         point in the glacier. Slower but more precise.
         """
         for i in axes(H̄, 1), j in axes(H̄, 2)
+            # We don't use apply_law! because we want to evaluate with custom inputs
```
JordiBolibar, line 119: I think that instead of using the function internally like this, it could be a good idea to create a new `apply_law` function with multiple dispatch to handle the type of custom inputs you want. It would definitely be cleaner.

src/models/target/target_D_pure.jl
```diff
@@ -96,6 +95,7 @@ function ∂Diffusivity∂H(
 
     # Derivative of the output of the NN with respect to input layer
     δH = 1e-4 .* ones(size(H̄))
+    # We don't use apply_law! because we want to evaluate with custom inputs
```
JordiBolibar, line 98: Same for all these.

JordiBolibar (commented):
Thanks for this work Facu! I'm aware this is work in progress, but nonetheless I added a couple of comments on things which could potentially be implemented differently (happy to discuss), and also to point out many places where docstrings and documentation will be really necessary.

src/ODINN.jl
```diff
@@ -70,6 +70,7 @@ const global root_plots::String = joinpath(root_dir, "plots")
 include(joinpath(root_dir, "src/setup/config.jl"))
 # Losses
 include(joinpath(root_dir, "src/losses/Losses.jl"))
+include(joinpath(root_dir, "src/losses/Regularizations.jl"))
```
albangossard, line 73: This is what makes the tests fail
```suggestion
include(joinpath(root_dir, "src/losses/Regularization.jl"))
```

PR #362, opened 2025-08-08
Introduce fast tests

albangossard:


JordiBolibar:
As discussed with @albangossard, the new protocol from now on will be:

- `main` is protected from direct pushes.
- All PRs need to go first to `dev`. This will be enforced during review of the PRs. We will first merge the PRs in `dev` with the light tests, and then we'll merge everything e.g. once per week into `main`, where the full test suite will be run.

JordiBolibar (approved):

Issue #365, opened 2025-08-12
Check gradient computed by the manual adjoint method when MB is enabled + fix it


albangossard:
Currently the contribution of the MB to the gradient is ignored in the manual adjoint. It's not clear whether in practice this contribution is important or not for inversions.
In order to check this hypothesis, we should check the gradient by comparing to the one obtained with SciMLSensivitity or FD.
The simulation period might have an impact on the accuracy as with longer simulations, MB has a higher impact.

Then we might want to add the contribution of MB to the gradient.

As suggested by @facusapienza21 we should make the computation of the MB contribution an option in the manual adjoint types.

albangossard:
In https://github.com/ODINN-SciML/ODINN.jl/pull/369 I implemented the MB contribution in the manual adjoints. We also have an option to discard the MB contribution by using for example `DiscreteAdjoint(VJP_method = DiscreteVJP(), MB_VJP=NoVJP())`.

## Comparison

Now we compare the results by running the same experiment as above with MB always enabled. We only vary the contribution of MB to the gradient computation.

### With MB contribution

#### Discrete adjoint
```
ratio  =  8.41e-03 < 0.01
angle  = -1.26e-05 < 1.0e-5
relerr =  9.72e-03 < 0.01
```
#### Continuous adjoint
```
ratio  =  1.36e-03 < 0.01
angle  = -1.26e-05 < 1.0e-5
relerr =  5.20e-03 < 0.01
```
### Without MB contribution

#### Discrete adjoint
```
ratio  = -1.37e-02 < 0.01
angle  = -1.26e-05 < 1.0e-5
relerr =  1.48e-02 < 0.01
```
#### Continuous adjoint
```
ratio  = -1.63e-02 < 0.01
angle  = -1.26e-05 < 1.0e-5
relerr =  1.73e-02 < 0.01
```

The difference of accuracy is rather limited and I expected something better.

PR #369, opened 2025-08-21
Add contribution of MB to gradient computation

albangossard:
Companion PRs:
- https://github.com/ODINN-SciML/Sleipnir.jl/pull/134
- https://github.com/ODINN-SciML/Muninn.jl/pull/32

Fixes https://github.com/ODINN-SciML/ODINN.jl/issues/365


JordiBolibar (commented):
No comments on my side, but I let @facusapienza21 revise the internals of the gradient computation.

src/inverse/SIA2D/gradient.jl
```diff
@@ -120,13 +132,13 @@ function SIA2D_grad_batch!(θ, simulation::FunctionalInversion)
                 ℓ += Δt[j-1]*ℓi
 
                 ### Custom VJP to compute the adjoint
-                λ_∂f∂H, dH_H = VJP_λ_∂SIA∂H(simulation.parameters.UDE.grad.VJP_method, λ[j], H[j], θ, simulation, t₀)
+                λ_∂f∂H, dH_H = VJP_λ_∂SIA∂H(simulation.parameters.UDE.grad.VJP_method, λ[j], H[j], θ, simulation, tj)
```
facusapienza21, line 135: Was this a bug in the original code? The fact that this was calling the time t0 rather that then current time? I think this probably had no effect before, since none of the dynamics depended of time before (with MB, this is no longer true)

src/inverse/SIA2D/gradient.jl
```diff
@@ -157,44 +169,63 @@ function SIA2D_grad_batch!(θ, simulation::FunctionalInversion)
...
+                    t = - integrator.t
+                    ∂ℓ∂H, ∂ℓ∂θ = backward_loss(
+                        loss_function,
+                        H_itp(t),
+                        H_ref_itp(t),
+                        t,
+                        glacier,
+                        θ,
+                        simulation;
+                        normalization = prod(N) * normalization
```
facusapienza21, line 200: @albangossard what is the logic behing doing this inside a `let` block?

src/inverse/SIA2D/gradient.jl
```diff
@@ -157,44 +169,63 @@ function SIA2D_grad_batch!(θ, simulation::FunctionalInversion)
...
+            cb_adjoint_MB = if simulation.parameters.simulation.use_MB
+                nSteps = Int(round((tspan[2]-tspan[1])/simulation.parameters.solver.step))
+                tstopsMB = - (tspan[1] .+ collect(1:nSteps) .* simulation.parameters.solver.step)
+                stop_condition_MB(λ, t, integrator) = Sleipnir.stop_condition_tstops(λ, t, integrator, tstopsMB)
+                DiscreteCallback(stop_condition_MB, effect_MB!)
+            else
+                CallbackSet()
             end
-            cb_adjoint_loss = DiscreteCallback(stop_condition, effect!)
+            cb = CallbackSet(cb_adjoint_MB, cb_adjoint_loss)
```
facusapienza21, line 221: Maybe worth mentioning somewhere that the tstops of these two callbacks don't need to be the same! I think it still works when MB and losses are aggregated at different time, but just worth checking this!

src/inverse/SIA2D/VJPs.jl
```diff
@@ -67,3 +68,41 @@ function VJP_λ_∂surface_V∂θ(VJPMode::DiscreteVJP, λx, λy, H, θ, simulat
...
+    MB = Enzyme.make_zero(H)
+    λH = deepcopy(λ) # Need to copy because Enzyme changes the backward gradient in-place
+    Enzyme.autodiff(
+        Reverse, MB_wrapper!, Const,
+        Duplicated(MB, λH),
+        Duplicated(H, λ_∂MB∂H),
+        Duplicated(simulation, _simulation),
+        Duplicated(glacier, _glacier),
+    )
+    return λ + λ_∂MB∂H
```
facusapienza21, line 103: I find a bit confusing that this returns the adjoint itself and the contribution of MB at the same time. I left another comment in `gradient.jl` regarding this point, but wouldn't make more sente to return just `λ_∂MB∂H` and then in the gradient calculation have a `.+=` instead of a `.=`?

src/inverse/SIA2D/gradient.jl
```diff
@@ -157,44 +169,63 @@ function SIA2D_grad_batch!(θ, simulation::FunctionalInversion)
...
-                integrator.u .= integrator.u .+ simulation.parameters.simulation.step .* ∂ℓ∂H
+                    integrator.u .= integrator.u .+ simulation.parameters.simulation.step .* ∂ℓ∂H
+                end
+            end
+            cb_adjoint_loss = DiscreteCallback(stop_condition_loss, effect_loss!)
+            effect_MB! = let simulation=simulation, glacier=glacier, H_itp=H_itp
+                function (integrator)
+                    t = - integrator.t
+                    λ_∂MB∂H = VJP_λ_∂MB∂H(simulation.parameters.UDE.grad.MB_VJP, integrator.u, H_itp(t), simulation, glacier, t)
+                    integrator.u .= λ_∂MB∂H
```
facusapienza21, line 210: Also following my comment before. I think it makes more sense to do `integrator.u .+= ` and just have the contribution of the adjoint. Am I missing something Alban?

src/inverse/VJPTypes.jl
```diff
@@ -34,4 +34,12 @@ Enzyme implementation of VJP used inside the adjoint calculation.
...
\ No newline at end of file
+end
+
+"""
+No VJP flavor when the contribution of a given term should not be computed inside the adjoint calculation (e.g. MB).
+
+`NoVJP`
+"""
+@kwdef struct NoVJP <: AbstractVJPMethod
+end
```
facusapienza21, line 45: I like it!

src/inverse/SIA2D/VJPs.jl
```diff
@@ -67,3 +68,41 @@ function VJP_λ_∂surface_V∂θ(VJPMode::DiscreteVJP, λx, λy, H, θ, simulat
...
+    λ_∂MB∂H = Enzyme.make_zero(H)
+    MB = Enzyme.make_zero(H)
+    λH = deepcopy(λ) # Need to copy because Enzyme changes the backward gradient in-place
+    Enzyme.autodiff(
+        Reverse, MB_wrapper!, Const,
+        Duplicated(MB, λH),
+        Duplicated(H, λ_∂MB∂H),
+        Duplicated(simulation, _simulation),
+        Duplicated(glacier, _glacier),
+    )
```
facusapienza21, line 102: I think is great this is super generic and it should work for any MB model, but notice that the adjoint contribution of the mass balance is actually very simple to compute and it is given by $\frac{\partial MB}{\partial H} \lambda$. For very simple MBs (like the ones we are using), the partial derivative $frac{\partial MB}{\partial H}$ is trivial to compute. This can be the default for some simple MBs models, but I imagine we can have at least one MB where we do this and we just use it for testing. Does this makes sense? Maybe it is a bit too much, but the actual contribution of MB to the adjoint is mathematically very simple, reason why I think we should consider having this done more "manually". What do you think @albangossard ?

facusapienza21 (commented):
@albangossard I left a few comments and questions. Overall, all these looks great! I just want to iterate on a few items I left in this PR.

Issue #366, opened 2025-08-12
Add initial condition inversion


albangossard:
We need to also invert the initial condition which is not currently supported in the code.

Goals of this issue:
- add support of the inversion with respect to H_0: optimizer, how we aggregate \theta with the different H_0, etc.
- update gradient computation: SciMLSensitivity + our manual adjoints
- quantify the impact of the time shift between the beginning of the simulation and the time where ice thickness data are available: should we shift the time at which ice thickness ground truth is defined?

albangossard:
Well I'm not sure we support IC inversion with SciMLSensitivity

albangossard:
Created #409

PR #370, opened 2025-08-21
Implementation of Inversion Pipeline wrt to Initial Condition of Ice Thickness

facusapienza21:
This PR addresses issue #366  regarding the incorporation of the optimization with respect to the initial condition of the glacier ice thickness. 

A few comments and questions:
- [ ] I took advantage of the Laws API to define the optimization with respect to H0. This is included as a "regressor", which makes sense, because H0 is a parameter of the trajectory and is a parameter we are regressing with respect to.
- [ ] I am not sure what is the best API for setting the training or not of the initial condition. My first guess was to include this as a boolean variable in `InversionParameters`, but then I realized is far easier to just add the regressor variable when defining the model (see code below). Of course, it does not make much sense to do both... My suggestion for now will be to leave as something we add in the `regressors` variable and maybe in the future have a nice API to automatically add this extra regressor when a flag is added. Thoughts?
```julia
if params.inversion.train_initial_conditions
    ic = InitialCondition(params, glaciers, :Farinotti2019Random)
else
    ic = ODINN.emptyIC()
end

model = Model(
    iceflow = SIA2Dmodel(params; A = LawA(nn_model, params)),
    mass_balance = nothing,
    regressors = (; A = nn_model, IC = ic)
    )
```
- [ ] Initial condition should be optimized with a regularization, which is very easy to add. The simplest way of doing this will be with the Regularization PR we already have as a draft. Maybe I can merge this here in order to have that running. 
- [ ] Add test to assess correctness of gradient with respect to initial condition.
- [ ] Implement this for more than one glacier. I am not sure what is the best strategy to incorporate the initial condition of multiple glaciers. This is because the component vector we are using to store all the variables needs to store vectors. For one single matrix associated to one single glacier, it is fine to store this a a vector than can later being recovered as a matrix (is it?). However, for multiple glaciers we need to store more stuff. What do you think is the best way of doing this? @albangossard any suggestion?
- [ ] Check implementation with Enzyme and SciMLSensitivity. So far, I have been doing this with the manual adjoints.


src/models/machine_learning/InitialConditions.jl
```diff
@@ -0,0 +1,48 @@
+export InitialCondition
```
JordiBolibar, line 1: I have a general comment regarding the term "initial condition". Do we refer by this as a list of potential initial conditions (e.g. ice thickness, surface velocities, glacier surface altimetry etc...) that we'd like to match? Or are we just talking about the ice thicknesses? Because if that's the case, I think we should rather call it Initial state.

src/parameters/InversionParameters.jl
```diff
@@ -16,12 +16,13 @@ A mutable struct that holds parameters for inversion processes. This struct is a
 """
 mutable struct InversionParameters{F<:AbstractFloat} <: AbstractParameters
     initial_conditions::Vector{F}
+    train_initial_conditions::Bool
```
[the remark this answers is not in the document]
JordiBolibar: Actually, the inversion wrt initial conditions should be made available for both workflows, so perhaps it should go in both types of simulations.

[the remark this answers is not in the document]
JordiBolibar: To me, this flag should probably go in `SimulationParameters` or even `UDEparameters`. I guess `UDEparameters` will still be used for a classical inversion, since we'll still need the adjoints, right?

src/simulations/functional_inversions/functional_inversion_utils.jl
```diff
@@ -139,13 +150,19 @@ end
...
+    save_every_iter::Bool = false,
+    logger::Union{<: TBLogger, Nothing} = nothing
+    ) where {AR <: Optimisers.AbstractRule}
 
     @info "Training with ADAM optimizer"
 
     # Create batches for inversion training
     simulation_train_loader = generate_batches(simulation)
 
+    # The variable θ includes all variables to being optimized, included initial conditions
```
JordiBolibar, line 165: "including"

JordiBolibar (commented):
Thanks Facu for initiating this work!

I agree with Alban that I'm not entirely convinced about the design of the workflow. See the review comments on how I think it might be more adequate to thread the flag. Moreover, if you already specify it in `regressors`, I'm not even sure we need that flag. 

In general, I would suggest adding more docstrings and comments, as well as updating the documentation with these new changes.

If I understand correctly, just by adding `H_0` as extra parameters in the `ComponentVector` `θ`, the workflow will already compute the necessary gradients of anything inside `θ`? I guess this should also be expanded for other initial conditions, such as e.g. a DEM or ice surface velocities?

JordiBolibar:
> @facusapienza21 it seems that we can use a `Matrix` inside a `ComponentArray`:
> 
> ```julia
> θ = zeros(4)
> H1 = zeros(4,4)
> H2 = zeros(5,5)
> v=ComponentArray(θ=θ,H1=H1,H2=H2)
> ```
> 
> Given a glacier the RGI ID is not always defined. However, once a glacier list has been determined it's fixed and we could map a `glacier_idx` to a symbol (or string) to be used to select the appropriate matrix in `v`. That way `v` is dynamically populated with the variables we need to initialize the ice thickness of each glacier and we don't have to store a `Vector{Matrix}` inside `v`. What do you think?

I'm not sure to get exactly what you mean. Would you mind showing a little snippet of the prototype you have in mind?

src/models/machine_learning/InitialConditions.jl
JordiBolibar: Hmm, it's a bit more nuanced than that. It depends on what observations you have at `t_0`. If you happen to have a DEM, then yes, you automatically know `S` AND your estimated `B`. And in that case, we should have a way to force or nudge the model to match `S` when inverting for `H`. This will give the most  accurate surface gradient and therefore the best `B`. For the vast majority of cases, inverting `H` or `B` is the same thing, because `B` will always be indirectly derived as `S - H`.

JordiBolibar:
I have an additional comment regarding an aspect we discussed in the last meeting and captured in #366. 

> quantify the impact of the time shift between the beginning of the simulation and the time where ice thickness data are available: should we shift the time at which ice thickness ground truth is defined?

How is this currently handled? I think we should have the flexibility to have the ground truth ice thickness any time during the simulated trajectory, not just at `t_0`. We should still use simulated ice thicknesses (e.g. Farinotti et al. (2019)) for the initial state, just to have something to start from. However, the snapshot used in the loss function to compute the error in `H` should be flexible, and thus `H` at the time where we have observations should be correctly stored during the forward run to be made available in the loss function.

facusapienza21:
Thank you for all the feedback @albangossard @JordiBolibar ! All good points and I knew it was a better idea to put this early for you to review rather than a more polish version. I will address these points shortly!

test/test_grad_loss.jl
```diff
@@ -87,26 +92,50 @@ function test_grad_finite_diff(
...
+        (; Y = nn_model)
+    elseif (target == :D_hybrid && train_initial_conditions)
+        (; Y = nn_model, IC = ic)
+    elseif (target == :D && !train_initial_conditions)
+        (; U = nn_model)
+    elseif (target == :D && train_initial_conditions)
+        (; U = nn_model, IC = ic)
+    else
+        nothing
+    end
```
albangossard, line 128: What do you think of using [MLStyle.jl](https://github.com/thautwarm/MLStyle.jl)?
This allows writing this big if/else as
```julia
regressors = @match (target, train_initial_conditions) begin
   (:A, false) => (; A=nn_model)
   (:A, true) => (; A=nn_model, IC=ic)
   ...
end
```
This package has minimal dependencies, so I think it's worth adding it!

src/models/machine_learning/ML_utils.jl
```diff
@@ -118,6 +118,7 @@ This function splits the glaciers and associated data in the `simulation` object
 - If the simulation results are empty, the function creates batches with empty results objects.
 """
 function generate_simulation_batches(simulation::FunctionalInversion)
+    # TODO: we need to change how this is done here, with a manual =1 batch size
```
albangossard, line 121: I'm opening a thread for this specific question:
> I am thinking the best way to do this may be modifying `build_simulation_batch` to split the component array with the IC of each glacier into each batch. Right now, the parameter vector \theta is shared among batches, and as long as these are just the NN coefficients is fine, but I don't think we want repeated the H0 of each single glacier in each single batch. I think it will be easier to have a \theta per batch where the NN coefficients are shared (as it is now) but where the IC component arrays is batched. This is one possible solution, let me know if you have other ideas @albangossard !

Yes I agree that this would be cleaner! Additionally, I started to implement the classical inversion and we'll have to add the `A` and `C` of each glacier to θ, so anyway this is a good practice to split the batch dependent parameters in that function.

test/runtests.jl
```diff
@@ -118,8 +118,19 @@ if GROUP == "All" || GROUP == "Core7"
     @testset "Inversion Tests w/ MB" inversion_test(use_MB = true, multiprocessing = false)
     @testset "Inversion Tests w/o MB w/ multiprocessing" inversion_test(use_MB = false, multiprocessing = true)
 end
+end
 
-@testset "Save results" save_simulation_test!()
+if GROUP == "All" || GROUP == "Core8"
```
albangossard, line 123: :warning: You need to add `Core8` and `Core9` to `ODINN/.github/workflows/CI_fast.yml`, otherwise these tests won't run in the _fast CI_.

test/test_grad_loss.jl
```diff
@@ -19,7 +20,17 @@ function test_grad_finite_diff(
...
-    rgi_ids = velocityLoss ? ["RGI60-11.03646"] : ["RGI60-11.03638"]
+    rgi_ids = if (velocityLoss && !multiglacier)
+        ["RGI60-11.03646"]
+    elseif (!velocityLoss && !multiglacier)
+        ["RGI60-11.03638"]
+    elseif (!velocityLoss && multiglacier)
+        ["RGI60-11.03638", "RGI60-11.01450"]
+    else
+        nothing
+    end
```
albangossard, line 31: A nice example where we could use MLStyle.jl :)

albangossard (commented):
I reviewed your recent commits and I think that you addressed most of the comments @facusapienza21 thanks!
We're close to having something that works. I agree with your suggestion of splitting theta in `generate_simulation_batches` which should make the whole thing work.

I'll review your whole PR one last time once it's ready for review.

facusapienza21:
@albangossard I addressed the points you mentioned. I fix everything and the code works, but I am struggling a bit with the logic for batching inside the gradient calculation (I opened #374 for this reason). It would be good to discuss this at some point soon, although I think what it is here makes the fix for now (take a look at how things are done to dispatch simulations). 

I noticed that adding IC slows down the optimization significantly, I will investigate this in the following days, but let me know if there is something that catches your attention.

src/inverse/SIA2D/gradient.jl
```diff
@@ -259,6 +268,14 @@ function SIA2D_grad_batch!(θ, simulation::FunctionalInversion)
                 throw("VJP method $(simulation.parameters.UDE.grad.VJP_method) is not supported yet.")
             end
 
+            # Contribution of initial condition to loss function
+            if haskey(θ, :IC)
+                λ₀ = sol_rev(-t₀)
+                # This contribution will come from the regularization on the initial condition
+                ∂L∂H₀ = 0.0
+                dLdθ.IC[Symbol("$(glacier.rgi_id)")] .+= λ₀ .+ ∂L∂H₀
```
albangossard, line 276: Concerning your comment

> I noticed that adding IC slows down the optimization significantly, I will investigate this in the following days, but let me know if there is something that catches your attention.

I open a thread here so that @JordiBolibar can jump into the discussion easily.
I see two potential explanations:
- 1) The use of `dLdθ.IC[Symbol("$(glacier.rgi_id)")]`, although it is type stable, adds a runtime dispatch which is more computationally intensive that just doing `dLdθ.IC[glacier_idx]`.
- 2) The duplication of the initial conditions before the parallelization in each process takes longer now that we have matrices. This is less likely since we are already copying the whole `simulation` struct which contains LOT of information.

test/runtests.jl
```diff
@@ -41,6 +41,10 @@ include("MB_VJP.jl")
 include("test_grad_loss.jl")
 include("save_results.jl")
 
+# Set random seed
+using Random
+Random.seed!(1234)
```
albangossard, line 46: Do you get reproducible results now?

PR #372, opened 2025-09-10
Code of conduct

facusapienza21:
Inspired by this [reading](https://third-bit.com/py-rse/teams.html#teams-coc), I think it would be ideal to include a code of conduct in the GH repository (and also in the docs!) as the project advances. I would like to reference this also in the docs, so we can think where to include it.


JordiBolibar (approved):
Thanks for this! I don't have anything to add. I agree with the main ideas behind it, although I have not been able to read it fully in detail :)

facusapienza21:
@JordiBolibar Are you ok if I made a few changes in the docs to include this and then merge?

PR #393, opened 2025-10-14
Regularization and Multi-Objective losses

facusapienza21:
This PR replaces #359 (pardon my Git). Reviews from the old PR #359 had been addressed here. 

Please take a look at how the regularization is implemented. Special comments and feedback is welcome regarding the definition of `MultiLoss`, which allows to custom different type of losses together (both empirical and regularization). 

A few points that I am a bit undecided or would need to think more are: 

- [ ] As it is, the hyperparameter of the loss is added as a tuple in `MultiLoss`. It may be better just to add an entry to each individual loss (a scaling factor)
- [ ] Some losses are evaluated continuously and other discrete... these last ones are a bit more tricky. I am not sure the best way of handling this right now, and in the most clean way. I would appreciate a future discussion on this! This is important since in reverse the gradients are evaluated in `t_nodes`, which do not include the initial and final timestep... I can try to change this too. Maybe we can define `AbstractContinuousLoss` and `DiscreteContinuousLoss` to dispatch them differently, and be able to handle both cases? I think this makes more sense (and coherent with the whole SciML desing), but I would like to think about this before moving forward. All comments are welcome @albangossard !
- [ ] I need to include more documentation and docstrings here


src/losses/MultiLoss.jl
```diff
@@ -0,0 +1,70 @@
+export MultiLoss
+export loss, backward_loss
+
+"""
+"""
+@kwdef struct MultiLoss{F <: AbstractFloat} <: AbstractLoss
+    losses::Union{Tuple{Vararg{<:AbstractLoss}}, Nothing} = (L2Sum(), )
```
albangossard: This is not type stable, which is definitely something we want for this specific part of the code!
If I understand properly, `MultiLoss{Sleipnir.Float}` isn't a concrete type because you don't parametrize the `<:AbstractLoss` term. A simpler way of implementing this would be to directly parametrize `MultiLoss` by {TL, TS}`. For example we would define
```julia
struct MultiLoss{TL, TS} <: AbstractLoss
    losses::TL
    λs::TS
end
```
and the concrete type of `TL` and `TS` would be determined in the inner constructor (see my next comment).

src/losses/Regularization.jl
```diff
@@ -0,0 +1,212 @@
...
+end
+
+function loss(
+    regType::TikhonovRegularization,
+    a::Matrix{F},
+    Δx::F,
+    Δy::F;
+    normalization::F=1.
+) where {F <: AbstractFloat}
+    return sum(∇²(a, Δx, Δy).^2.0)
```
albangossard: To follow up on that comment https://github.com/ODINN-SciML/ODINN.jl/pull/359#discussion_r2427032406 for me Tikhonov regularization has the form $\|L x\|_2^2$ with $L$ any linear operator but I know this is very community dependent. In image correlation for mechanics for example when people talk about Tikhonov regularization, they frequently mean $L=Id$.

I think it would be great to be able to customize this, or at least — if we don't want to implement this right now — change the name so that it's not restricted to a special case. What do you think?

src/losses/Regularization.jl
```diff
@@ -0,0 +1,212 @@
...
+) where {F <: AbstractFloat}
+
+    Δx, Δy = glacier.Δx, glacier.Δy
+
+    if !isnothing(simulation.model.machine_learning)
+        simulation.model.machine_learning.θ = θ
+    end
+    Vx, Vy, V = Huginn.V_from_H(simulation, H, t, θ)
+
+    mask = is_in_glacier(H, lossType.distance) .& (V .> 0.0)
```
albangossard: `lossType` isn't defined here

JordiBolibar:
I have a general question regarding the application of the loss at discrete or specific timesteps.  For a case where we have surface velocity observations at different timesteps of the trajectory, how is this being handled? We evaluate at all time steps and the step is ignored if there are no obs to compare to? I guess there must be some similarities in terms of how to handle this with what is being discussed here for the initial conditions, right?

albangossard (commented):
> For a case where we have surface velocity observations at different timesteps of the trajectory, how is this being handled? We evaluate at all time steps and the step is ignored if there are no obs to compare to?

This depends of which adjoint you're referring to. For the moment in the **discrete adjoint** we compute the predicted ice surface velocity every month and we compare to the observation that we have. We apply a filter to keep the observation within a time window of a few days. If there is no observation we ignore that time step.
In the **continuous adjoint** the quadrature gives you a few points where to compute the loss (whatever LossH, LossV or LossHV), and then you retrieve the velocity observation with the same time window as for the discrete adjoint.

Now you see the very important point: depending on the quadrature points and on the quality of your velocity data (dense or not), you might be missing most of the glacier dynamics!

JordiBolibar:
> Now you see the very important point: depending on the quadrature points and on the quality of your velocity data (dense or not), you might be missing most of the glacier dynamics!

Then I wonder, is there a way to "enforce" that the forward solve hits some timesteps? Even at the cost of more computational time due to smaller timesteps? Because in the summer season, missing the target by a few days can imply that there can be a big mismatch between model and observations (e.g. if we take into account the surface melt of the previous few days).

src/losses/Regularization.jl
facusapienza21: I have changed this is the last version to provide this flexibility!

facusapienza21:
@albangossard @JordiBolibar we can discuss this points in our next meeting, but I don't think is too difficult to solve it. My preference is still to use the continuous adjoint with a dense interpolation, which is what we are kind of already using.

src/losses/Regularization.jl
```diff
@@ -0,0 +1,212 @@
...
+    ∂a∂y_dual = Huginn.avg_x(∂a∂y)
+    # Second derivative
+    ∂2a∂x2_dual = Huginn.diff_x(∂a∂x_dual, Δx)
+    ∂2a∂y2_dual = Huginn.diff_y(∂a∂y_dual, Δy)
+    # Evaluate in primal grid
+    ∂2a∂x2 = Huginn.avg_y(∂2a∂x2_dual)
+    ∂2a∂y2 = Huginn.avg_x(∂2a∂y2_dual)
+
+    ∇²a = zero(a)
+    Huginn.inn(∇²a) .= ∂2a∂x2 .+ ∂2a∂y2
```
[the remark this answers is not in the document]
facusapienza21: I proposed a solution in one of the last commits, but not sure is ideal:
```julia
nx, ny = size(a)
∇²a = ∂2a∂x2 .+ ∂2a∂y2
∇²a_inner = [i == 1 || i == nx || j == 1 || j == ny ? 0.0 : ∇²a[i - 1, j - 1] for i = 1:nx, j=1:ny]
return ∇²a_inner
```

PR #395, opened 2025-10-15
Customize the VJP functions in the laws

albangossard:
PR associated to:
- https://github.com/ODINN-SciML/Sleipnir.jl/pull/145
- https://github.com/ODINN-SciML/Muninn.jl/pull/41
- https://github.com/ODINN-SciML/Huginn.jl/pull/115

Things that I still have to do:
- [x] Update the VJP documentation
- [x] Add a new page about the VJP customization of the laws
  * [x] Simple VJP customization
  * [x] Example of how to define your own cache
- [x] Restructure the law code by
  * [x] Splitting things into different files, waiting https://github.com/ODINN-SciML/Sleipnir.jl/pull/146
  * [x] Cleaning the constructors, waiting https://github.com/ODINN-SciML/Sleipnir.jl/pull/146
  * [x] Removing the `scalar` argument, waiting https://github.com/ODINN-SciML/Sleipnir.jl/pull/146
- [x] Update the show functions to include information about the VJPs
- [x] Make a cleaner implementation of `prepare_vjp_law` for `FunctionalInversion`


docs/src/vjp_laws.jl
```diff
@@ -0,0 +1,377 @@
...
+# ## Explanations
+
+# ### High-level summary
+
+# At the user level the customization can be made by implementing hand-written VJPs through the following functions:
+# - `f_VJP_input!(...)` — VJP w.r.t. inputs
+# - `f_VJP_θ!(...)` — VJP w.r.t. parameters θ
+# - You may also implement your own precompute function to cache expensive computations which is the purpose of `p_VJP!(...)`. This function is called before solving the adjoint iceflow PDE.
+
+# Internally when the user does NOT provide VJPs, ODINN uses a default AD backend (via [DifferentiationInterface.jl](https://github.com/JuliaDiff/DifferentiationInterface.jl)).
```
JordiBolibar: Just to be clear, does this work already with DifferentiationInterface? I understood that it would take a while to adapt all the interfaces.

docs/src/vjp_laws.jl
```diff
@@ -0,0 +1,377 @@
...
+#   ```
+#   law_VJP_input(law::AbstractLaw, cache, simulation, glacier_idx, t, θ)
+#   ```
+#   - Intent and behavior:
+#     - These are called during the adjoint solve to compute parameter and input VJPs for the law at time `t` and for parameters `θ`.
+#     - They can either compute the VJPs directly or use cached VJP information that has been already computed in the user-supplied `p_VJP!` VJP function. The `cache` allows storing useful information from the forward or from the precomputation step.
+#     - They therefore carry the runtime context (simulation, glacier index, time, θ) which is necessary for adjoint calculations.
+
+# ### Workflow
+# How the pieces compose in practice?
```
JordiBolibar:
```suggestion
# How do the pieces compose in practice?
```

docs/src/vjp_laws.jl
```diff
@@ -0,0 +1,377 @@
...
+# - If you, as a user, provide custom VJP functions (through `f_VJP_input!`/`f_VJP_θ!`, or through `p_VJP!`), ODINN will use them directly at adjoint time and will skip the AD fallback path. You can also provide your own precompute wrapper and cache to optimize expensive computations.
+# - If you do NOT provide VJP functions, ODINN runs the AD fallback:
```
JordiBolibar: How robust is this? Can we say it can generalize to multiple laws?

Something I have not seen is where in the ODINN API we can change the AD backend of DI.

docs/src/vjp_laws.jl
```diff
@@ -0,0 +1,377 @@
...
+#   and
+#   ```
+#   law_VJP_input(law::AbstractLaw, cache, simulation, glacier_idx, t, θ)
+#   ```
+#   - Intent and behavior:
+#     - These are called during the adjoint solve to compute parameter and input VJPs for the law at time `t` and for parameters `θ`.
+#     - They can either compute the VJPs directly or use cached VJP information that has been already computed in the user-supplied `p_VJP!` VJP function. The `cache` allows storing useful information from the forward or from the precomputation step.
+#     - They therefore carry the runtime context (simulation, glacier index, time, θ) which is necessary for adjoint calculations.
+
+# ### Workflow
```
JordiBolibar: Something I would make even more explicit here, taking into account that a wide audience might read this is:

- Explain that for advanced users seeking maximum performance and control, custom VJPs are the way.
- For users not wanting to deal with these internals, use the AD option to automatically have it computed for you. 

By explicitly saying to people who might be confused by all the jargon here that it is OK to go AD, basically we are lowering the bar for those users to work with ODINN, which is also what we want. It is great that we can cater both advanced and less experienced users in this front.

docs/src/vjp_laws.jl
```diff
@@ -0,0 +1,377 @@
...
+# - `vjp_inp`: a placeholder to store the result of the VJP with respect to the inputs, must be of a type that matches the one the inputs are defined in
+# In order to know the type of the inputs, simply run `generate_inputs(law.f.inputs, simulation, glacier_idx, t)`.
+
+# ### Using the preparation object
+
+# !!! error
+#     For the moment, using the preparation object at the user level is not supported yet.
+
+# ### Best practices and debugging tips
+# - If you supply custom VJPs, test them with finite-difference checks for both inputs and parameters. ODINN does not check that the correctness of your implementation!
```
JordiBolibar: Here I would perhaps provide a link to a tests doing so, as an example.

docs/src/vjp_laws.jl
```diff
@@ -0,0 +1,377 @@
...
+)
+
+# !!! success
+#     This law is applied only once before the beginning of the simulation, and the VJP are precomputed using our own implementation.
+
+
+# ## Simple cache customization
+
+# In this last section we illustrate how we can define our own cache to store additional information.
+# Our use case is the interpolation of the VJP on a coarse grid.
```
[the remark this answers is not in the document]
JordiBolibar: I guess this is what Facu was playing with, in order to accelerate the computations? When you mean coarse grid, you mean that the VJP is just computed for some grid points of the glacier, and then you interpolate between them?

scripts/law_plots.jl
```diff
@@ -81,8 +81,13 @@ prediction = generate_ground_truth_prediction(glaciers, params, model, tstops)
 
 ### Figures
 
-plot_law(prediction.model.iceflow.C, prediction, law_inputs, 1, nothing)
+ODINN.connect_electron_backend() # If you want to visualize the plots below properly on Ubuntu 24.04
```
JordiBolibar: This sounds so retro 😆

albangossard: This was super annoying to make it work on my laptop...

docs/src/vjp_laws.jl
albangossard: Good idea thanks! I added a paragraph at the beginning of this section mentioning what you said.

docs/src/vjp_laws.jl
albangossard: Oops thank you
c5b2cef

PR #413, opened 2025-10-25
Implementation of D Law based on Interpolation Cache

facusapienza21:
*** Please still don't review ***

Work in progress regarding the implementation of the interpolation cache in order to run complex inversion pipelines that require multiple AD evaluations of the NN. 

Task to be completed before merge

- [x] `init_cache` needs to be provided even when Interpolation is not being used
- [ ] Right now, we define a function `feed_input_cache!` to be able to update the knots of the intepolation for the later evaluation in `p_VJP!`. It is unclear this is the best logic to implement this, but right now this cannot be done inside the `Law` declaration since in that scope we no longer have access to `simulation` and `results`. To be determined how to bypass this in conversation with @albangossard 

Companion PRs:
- https://github.com/ODINN-SciML/Huginn.jl/pull/123


src/laws/Laws.jl
```diff
@@ -82,8 +82,11 @@ U_law = LawU(nn_model, params; max_NN = 50.0, prescale_bounds = [bounds_H, bound
 function LawU(
     nn_model::NeuralNetwork,
     params::Sleipnir.Parameters;
-    max_NN::Union{F, Nothing} = 50.0,
-    prescale_bounds::Union{Vector{Tuple{F,F}}, Nothing} = [(0.0, 300.0), (0.0, 0.5)],
+    max_NN::Union{F, Nothing} = nothing,
+    prescale_bounds::Union{Vector{Tuple{F,F}}, Nothing} = nothing,
+    precompute_interpolation::Bool = true,
+    precompute_VJPs::Bool = true,
+    dummy_float::F = 1.0, # This is just to recognize the type F when no float argument is provided. Not always, but sometimes this produces an error of "F not been defined"... I am sure there is a better solution to this but I couldn't find it. Doing this for now.
```
facusapienza21: @albangossard @JordiBolibar do you have any suggested solution for this? When I pass the possible `F` arguments as `nothing` (in this case, `prescale_bounds` and `max_NN`) I get sometimes (but not always, which is weird and annoying) an error since the type `F` cannot be found by the arguments (of course, since there are no float). Do you know how to fix this? I just introduced a dummy float variable so `F` is defined by default, just to be able to run this code for now.

albangossard: I think this will solve your problem
```suggestion
    max_NN::Union{<: AbstractFloat, Nothing} = nothing,
    prescale_bounds::Union{Vector{Tuple{<: AbstractFloat,<: AbstractFloat}}, Nothing} = nothing,
    precompute_interpolation::Bool = true,
    precompute_VJPs::Bool = true,
```

test/test_grad_loss.jl
```diff
@@ -187,20 +210,31 @@ function test_grad_finite_diff(
...
+                        non_zero = M .> 1.0
+                        idxs = rand(findall(non_zero), max_params)
+                        mask = falses(size(M)...)
+                        mask[idxs] .= 1
+                        key = Symbol("$(glacier.rgi_id)")
+                        θ_mask.IC[key] .= mask
+                    end
+                else
+                    # Mask parameter vector
+                    indx = ODINN.sample(1:length(θ.U), max_params; replace = false)
```
albangossard: I find this kind of hack super dangerous. What happens then if we want to test the gradient with respect to other invertible parameters? (this is what I implemented in my PR)
Depending on the randomness we might not be testing the gradient with respect to some of the parameters.

I would be ok to use this hack to speed-up the computation but only on the initial condition.
What do you think?

src/laws/Laws.jl
facusapienza21: Seems to be working for now!!! Thank you!

test/test_grad_loss.jl
```diff
@@ -187,20 +252,35 @@ function test_grad_finite_diff(
 
         ### Further computes derivatives with FiniteDifferences.jl (stepsize algorithm included)
 
-        if train_initial_conditions
-            # TODO: Unify this with definition of f and mask
+        if n_params > max_params
```
albangossard: As discussed during the last meeting I strongly oppose to make this kind of behavior the default one. It's ok to test only a subset of parameters of the IC, but not of the whole theta. Imagine that we test the gradient in the case of IC inversion + inversion of a single A per glacier. The gradient wrt A may not be tested.

src/models/target/target_utils.jl
```diff
@@ -192,15 +192,36 @@ function ComponentVector2Vector(cv::ComponentVector)
...
+function create_interpolation(A::Vector; n_interp_half::Int, dilation_factor = 1.0)
+    A_interp_unif = LinRange(0.0, dilation_factor * maximum(A), n_interp_half) |> collect
     pos_values = A[A .> 0.0]
     if length(pos_values) > 0
-        A_interp_quantiles = quantile!(A[A .> 0.0], LinRange(0.0, 1.0, n_interp_half))
+        # We remove the 0.0 and 1.0 quantile since these are included already in the
+        # uniform interpolation
+        quantile_range = LinRange(0.0, 1.0, n_interp_half + 2)[begin + 1: end - 1]
+        A_interp_quantiles = quantile!(A[A .> 0.0], quantile_range)
         A_interp = vcat(A_interp_unif, A_interp_quantiles)
```
albangossard: Is the interpolation gonna work if the two vectors contain redundant values?
```suggestion
        A_interp = unique(vcat(A_interp_unif, A_interp_quantiles))
```

src/models/target/target_D_pure.jl
```diff
@@ -71,55 +71,55 @@ function Diffusivity(
     H̄, ∇S, θ, simulation, glacier_idx, t, glacier, params
     )
     iceflow_cache = simulation.cache.iceflow
-    U = iceflow_cache.U.value
+    iceflow_model = simulation.model.iceflow
+
+    # TODO: this can be replace by interpolation too!
+    U = iceflow_model.U.f.f(iceflow_cache.U, (; H̄ = H̄, ∇S = ∇S), θ)
```
albangossard: This completely breaks the design of the law API! Here you should use `U = iceflow_cache.U.value` because this has already been computed in `apply_all_non_callback_laws!`. This allows accessing the prediction of a law, no matter if it's a callback law or not.

src/models/target/target_D_pure.jl
```diff
@@ -198,124 +178,50 @@ function ∂Diffusivity∂θ(
...
+"""
 function Diffusivityꜛ(
     target::SIA2D_D_target;
     H̄, ∇S, θ, simulation, glacier_idx, t, glacier, params
     )
-    iceflow_cache = simulation.cache.iceflow
-    U = iceflow_cache.U.value
-    return U
+    return Diffusivity(
+        target;
```
albangossard: Same thing here, you should use the cache.

src/laws/Laws.jl
```diff
@@ -93,21 +95,68 @@ function LawU(
...
+            zeros(nx - 1, ny - 1),
+            # zeros(nx - 1, ny - 1),
+            # zeros(nx - 1, ny - 1, length(θ)),
+            H_nodes,
+            H_nodes,
+            grad_itp
+            )
+    end
+
+    p_VJP! = let nn_model = nn_model
```
albangossard:
```suggestion
    p_VJP! = let smodel = smodel, prescale = prescale, postscale = postscale
```

albangossard (changes requested):
There are a few things to change, especially how you interact with the cache. I know that in practice you'll never use callback laws for the pure diffusivity inversion, but I think it's important to keep a coherent API across the different targets.

facusapienza21:
@albangossard maybe we can discuss this points next week. I am a bit confused about how to interact with the cache here.

albangossard:
@facusapienza21 I fixed the gradient computation. Currently the code takes some time to execute since we don't interpolate the VJP with respect to H and ∇S. I'll try to implement it this afternoon.

By the way I opened an issue https://github.com/ODINN-SciML/ODINN.jl/issues/423 to rename the "ꜛ" functions to something that makes more sense.

facusapienza21:
Yes! Let's discuss all this tomorrow. I would like to keep the custom NN test (at least one) since I noticed in many cases the simple NN tests work, but some custom NN breaks everything (which took me a lot to realize this, since the tests were passing but my code was failing. We can discuss this tomorrow!)

PR #418, opened 2025-11-04
Split discrete VJPs into two seperate functions

albangossard:
We now have very similar runtimes between the discrete and continuous VJPs:
image


facusapienza21 (approved):
Nice!!! Much better this way, and happy to see a bit of improvement in performance!

Issue #416, opened 2025-10-30
Make forward simulation stops at relevant tstops based on data


albangossard:
Currently we use an explicit Euler scheme with a fixed time step to solve the discrete adjoint. Besides the accuracy of the generated solution, this constraints a lot how we integrate the ice surface velocity into the loss. We should instead use sciml and compute the solution at required time steps so that this matches where we have available data.

albangossard:
Following yesterday's discussion, this issue doesn't make sense. Instead we want to make sure that the forward simulation stops at relevant tstops according to the data we have. I'll update its title

PR #414, opened 2025-10-29
Implement classical inversion

albangossard:
Closes https://github.com/ODINN-SciML/ODINN.jl/issues/339

The most important changes are in the first commit. The second commit just consists in renaming `FunctionalInversion` by `Inversion`.
I didn't rename the name of the files since we have ongoing PRs. I created a dedicated issue for that: https://github.com/ODINN-SciML/ODINN.jl/issues/415

Associated PR: https://github.com/ODINN-SciML/Huginn.jl/pull/125

PR #420, opened 2025-11-12
Adaptive tstops

albangossard:
Since I needed the changes of https://github.com/ODINN-SciML/ODINN.jl/pull/414 to address https://github.com/ODINN-SciML/ODINN.jl/issues/416, I propose to merge this PR in `classicalInv`

Associated PRs:
- https://github.com/ODINN-SciML/Sleipnir.jl/pull/150
- https://github.com/ODINN-SciML/Muninn.jl/pull/43
- https://github.com/ODINN-SciML/Huginn.jl/pull/128


src/losses/Losses.jl
```diff
@@ -246,64 +246,66 @@ end
...
     θ,
     simulation,
     normalization::F,
+    Δt,
 ) where {F <: AbstractFloat}
-    lH = loss(lossType.loss, H_pred, H_ref, normalization)
-    return lH
+    if isnothing(H_ref)
+        return 0.0
+    else
```
facusapienza21: Isn't a bit dangerous to have this here? If we call for the evaluation of a loss with `LossH` we should always provide with `H_ref`. I guess you are trying to be able to make this function call when the t where you stop does not have ice thickness data (but maybe it does have velocity data), but don't you think this can be included in other part of the code? If not, I will suggest having a nice comment here explaining why this `if isnothing(H_ref)`.

src/losses/Losses.jl
```diff
@@ -313,47 +315,39 @@ function loss(
...
-    # Scale loss function 
+    # Scale loss function
     ℓ_scale = if lossType.scale_loss
         ℓ / mean(Vx_ref[mask].^2 .+ Vy_ref[mask].^2)^0.5
     else
         ℓ
     end
 
-    return ℓ_scale
+    return ℓ_scale*Δt.V
```
facusapienza21:
```suggestion
    return ℓ_scale * Δt.V
```

src/inverse/SIA2D/gradient.jl
```diff
@@ -61,34 +68,63 @@ function SIA2D_grad_batch!(θ, simulation::Inversion)
...
         H = result.H
         glacier = simulation.glaciers[i]
 
-        # Reference data
-        t_ref = glacier.thicknessData.t
-        H_ref = glacier.thicknessData.H
+        ## 2- Reference data
+
+        # Discretization for the ice thickness loss term
+        tH_ref = tdata(glacier.thicknessData)
```
facusapienza21: I will add a comment that adding this here does not harm even when there is no thicknessData, since this will be nothing in that case, right? Just for clarity in the code

facusapienza21 (changes requested):
(I haven't yet finished the review! But I will submit so I am sure the comments are saved)

src/inverse/SIA2D/gradient.jl
```diff
@@ -97,22 +133,42 @@ function SIA2D_grad_batch!(θ, simulation::Inversion)
             # Define empty object to store adjoint in reverse mode
             λ  = [zero(result.B) for _ in 1:k]
 
-            res_backward_loss = map(j -> backward_loss(
+            res_backward_loss = map(1:k) do j
+                tj = t[j]
+                indThickness = findfirst(==(tj), tH_ref)
+                indVelocity = findfirst(==(tj), tV_ref)
+                Δtj = (;
```
facusapienza21: See comment here about handling discrete and continuous losses separately

src/inverse/SIA2D/gradient.jl
```diff
@@ -125,31 +181,37 @@ function SIA2D_grad_batch!(θ, simulation::Inversion)
...
-
-                # Contribution to the loss
-                dLdθ .+= Δt[j-1] .* (isnothing(∂ℓ∂θ) ? λ_∂f∂θ : λ_∂f∂θ .+ ∂ℓ∂θ)
+                ### Contribution to the loss
+                # For ∂ℓ∂θ, time discretization is already included in the loss, so no need to multiply by Δt
+                dLdθ .+= Δt[j-1] * λ_∂f∂θ .+ something(∂ℓ∂θ, 0.0)
             end
 
+            ### Check consistency between forward and reverse
+            @assert ℓ ≈ loss_per_glacier[i] "Loss in forward and reverse do not coincide: $(ℓ) != $(loss_per_glacier[i])"
```
facusapienza21: Replace by `isapprox`

src/inverse/SIA2D/gradient.jl
```diff
@@ -242,19 +317,30 @@ function SIA2D_grad_batch!(θ, simulation::Inversion)
             λ₁ = zero(H[end])
 
             # Include contribution of loss from last step since this is not accounted for in the discrete callback
-            if tspan[2] ∈ t_ref
+            if tspan[2] ∈ tstops
```
facusapienza21: Is there any way with the new logic with the timestepts to remove the need from this code and include all together?

src/inverse/SIA2D/gradient.jl
```diff
@@ -61,34 +68,63 @@ function SIA2D_grad_batch!(θ, simulation::Inversion)
...
-        @assert length(H) == length(H_ref)
-        @assert size(H[begin]) == size(H_ref[begin])
+        # Discretization provided to the loss as a named tuple with the discretization for each term
+        Δt_HV = (; H=ΔtH, V=ΔtV)
+
+        ## 3- Determine tstops in the same way as what is done in the forward and check that this matches
+        tstops = Huginn.define_callback_steps(params.simulation.tspan, params.solver.step)
+        tstops = sort(unique(vcat(tstops, params.solver.tstops, tH_ref, tV_ref)))
+
+        @assert t ≈ tstops "Times in tstops and reference times in result do not coincide."
```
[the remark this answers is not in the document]
albangossard: I don't think we expect an error on the time steps. I tried to replace by a `==` and the tests are passing, so I suggest that we replace by an exact comparison here

src/losses/Losses.jl
```diff
@@ -246,64 +246,66 @@ end
...
-    # 2- Compute the predicted velocity Vx_pred, Vy_pred, V_pred
+    # Compute the predicted velocity Vx_pred, Vy_pred, V_pred
     if !isnothing(simulation.model.machine_learning)
         simulation.model.machine_learning.θ = θ
     end
     Vx_pred, Vy_pred, V_pred = Huginn.V_from_H(simulation, H_pred, t, θ)
     # TODO: in the future we should dispatch wrt the iceflow model
 
-    mask = is_in_glacier(H_ref, lossType.loss.distance) .& (V_ref .> 0.0)
+    mask = is_in_glacier(H_pred, lossType.loss.distance) .& (V_ref .> 0.0)
```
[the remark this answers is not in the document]
albangossard: Good point thank you! I did this because we don't necessarily have a valid `H_ref` at some time steps in the ice surface velocity loss.
I'll implement what you suggest in an upcoming commit

albangossard (commented):
Thanks for this very complete review @facusapienza21!!

I won't have the time to address all your comments today, but I replied to the most critical ones, especially the ones where I don't get what you suggest to implement.
Feel free to complement the review and I would be happy to implement cleaner implementations once I get your points!

src/losses/Losses.jl
facusapienza21: Sounds good!

src/inverse/SIA2D/gradient.jl
```diff
@@ -61,34 +68,63 @@ function SIA2D_grad_batch!(θ, simulation::Inversion)
...
+        # Let's compute the forward loss inside the gradient computation for verification purpose
+        ℓ = 0.0
+
         apply_all_callback_laws!(simulation.model.iceflow, simulation.cache.iceflow, simulation, i, tspan[2], θ)
         precompute_all_VJPs_laws!(simulation.model.iceflow, simulation.cache.iceflow, simulation, i, tspan[2], θ)
 
         if typeof(simulation.parameters.UDE.grad) <: DiscreteAdjoint
 
+            if useThickness && useVelocity && tH_ref != tV_ref
+                @warn "Correctness of the gradient is not guaranteed when using the discrete adjoint with different time steps for H and V terms! Use the continuous adjoint instead."
```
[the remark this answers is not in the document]
facusapienza21: mmm I will need more context to understand this, but happy to discuss what can be causing this!

src/losses/Losses.jl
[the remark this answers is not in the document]
albangossard: As discussed on slack, in fa2d0839d417977be7da41d30df2706f0fd7cf17 I simplified the mask computation by taking into account only positive velocities. We now assume that the reference ice surface velocity has been preprocessed and cleaned. Maybe it's worth discussing this during our next meeting!

PR #435, opened 2025-11-25
Merge dev into main

albangossard:


albangossard:
> All good! Please don't squash this PR so we don't lose the history of the last developments!

Yes of course!
I'm waiting the macOS tests to pass

Issue #387, opened 2025-09-25
Improve tests with Aqua.jl


albangossard:
https://github.com/JuliaTesting/Aqua.jl

PR #451, opened 2025-12-20
Aqua tests added

JordiBolibar:
Adding Aqua tests to the test suite. So far, I've added all of them except for:

- [Method ambiguities](https://juliatesting.github.io/Aqua.jl/stable/ambiguities/), which has conflicts with the use of `Union{F, Nothing}`.
- [Undocumented names](https://juliatesting.github.io/Aqua.jl/stable/undocumented_names/), which enforces documenting absolutely everything. For now it's set to `broken=true`.

Addresses https://github.com/ODINN-SciML/ODINN.jl/issues/387

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -34,6 +34,7 @@ using JLD2
 using Plots
 using Infiltrator
 using OrdinaryDiffEq
+using LinearAlgebra
 using Optim
 using SciMLSensitivity
 using Statistics
--- a/test/test_grad_loss.jl
+++ b/test/test_grad_loss.jl
@@ -63,7 +63,7 @@ function test_grad_finite_diff(
         @assert target == :A "When testing classical inversion, only target A is supported"
     end
 
-    print("> Testing target $(target) with adjoint $(adjointFlavor) and loss $(Base.typename(typeof(loss)).name)")
+    print("> Testing target $(target) with $(adjointFlavor) and $(Base.typename(typeof(loss)).name)")
     println(use_MB ? " and with MB" : "")
 
     # Determine if we are working with a velocity loss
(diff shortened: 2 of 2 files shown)


src/laws/auto_VJP.jl
```diff
@@ -82,41 +82,41 @@ function prepare_vjp_law(
     return VJPsPrepLaw(f_θ_first, f_inp_first, prep_θ, prep_inp)
 end
 
-function ∂law∂inp!(backend::DI.AutoZygote, vjpsPrepLaw::AbstractPrepVJP, inp, θ)
+function ∂law∂inp!(vjpsPrepLaw::AbstractPrepVJP, backend::DI.AutoZygote, inp, θ)
```
albangossard: Why did you change the signature of these methods? This breaks the API and you'll need to change this accordingly in the documentation and in Sleipnir as well.

Project.toml
```diff
@@ -17,14 +17,12 @@ Downloads = "f43a241f-c20a-4ad4-852c-f6b1247861c6"
 Enzyme = "7da242da-08ed-463a-9acd-ee780be4f1d9"
 EnzymeCore = "f151be2c-9106-41f4-ab19-57ee4f262869"
 FastGaussQuadrature = "442a2c76-b920-505d-bb47-c5924d526838"
-FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
```
albangossard: I think that with the fast tests, we run in the default environment and hence we need FiniteDifferences and JET there.

test/runtests.jl
```diff
@@ -126,9 +128,9 @@ ENV["GKSwstype"] = "nul"
                 ContinuousAdjoint(VJP_method = ContinuousVJP()); thres = [2e-2, 1e-5, 2e-2])
             @testset "Continuous adjoint with Enzyme VJP vs finite differences" test_grad_finite_diff(
                 ContinuousAdjoint(VJP_method = ODINN.EnzymeVJP());
-                thres = [5e-4, 2e-8, 5e-4])
+                thres = [5e-4, 7e-7, 2e-3])
             @testset "SciMLSensitivity adjoint with Enzyme VJP vs finite differences" test_grad_finite_diff(
-                ODINN.SciMLSensitivityAdjoint(); thres = [5e-4, 5e-8, 5e-4])
+                ODINN.SciMLSensitivityAdjoint(); thres = [5e-4, 7e-7, 2e-2])
```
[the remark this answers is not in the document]
JordiBolibar: Yes, although if you look at the other thresholds above (not changed in this PR), some already have those values.

PR #452, opened 2026-01-08
Introduces mask from glacier into loss V

facusapienza21:
Compation PR:
- https://github.com/ODINN-SciML/Sleipnir.jl/pull/158

### DIFF
--- a/src/simulations/inversions/callback_utils.jl
+++ b/src/simulations/inversions/callback_utils.jl
@@ -80,7 +80,7 @@ function callback_diagnosis(θ, l, simulation; save::Bool = false,
             improvement = (l - simulation.results.stats.losses[end - step]) /
                           simulation.results.stats.losses[end - step]
         end
-        printProgressLoss(iter,
+        printProgressLoss(length(simulation.results.stats.losses),
             simulation.results.stats.niter, l, improvement)
     end
 


src/losses/Losses.jl
```diff
@@ -314,7 +314,11 @@ function loss(
     Vx_pred, Vy_pred, V_pred = Huginn.V_from_H(simulation, H_pred, t, θ)
     # TODO: in the future we should dispatch wrt the iceflow model
 
-    mask = (V_ref .> 0.0)
+    mask = if length(simulation.glaciers[1].mask_loss) > 0
+        simulation.glaciers[1].mask_loss
+    else
+        V_ref .> 0.0
```
albangossard: In which situation do we fall in this case? From [Sleipnir's code](https://github.com/ODINN-SciML/Sleipnir.jl/pull/158/files#diff-28c1b2a5801fbf747667baa652b500b1ef83dc2b04e659d3ef8f9af242fb0292R292-R296) I see that `mask_loss` is always initialized with a non empty matrix, isn't it?

src/simulations/inversions/callback_utils.jl
```diff
@@ -65,9 +65,12 @@ function callback_diagnosis(θ, l, simulation; save::Bool = false,
...
 
-    push!(simulation.results.stats.losses, l)
-    push!(simulation.results.stats.θ_hist, θ.u)
-    push!(simulation.results.stats.∇θ_hist, θ.grad)
+    iter = θ.iter
+
+    # We save a copy of the state parameters as these are stored as a reference
+    push!(simulation.results.stats.losses, copy(l))
+    push!(simulation.results.stats.θ_hist, copy(θ.u))
+    push!(simulation.results.stats.∇θ_hist, copy(θ.grad))
```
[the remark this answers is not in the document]
facusapienza21: This is the reason why the gradients were being saved all the same!

src/losses/Losses.jl
facusapienza21: I was thinking in the case we manually construct the glacier object by hand (e.g., this is what I am doing with Halfalr solutions), so I though it was convenient to have a behaviour when we pass an empty BitMatrix. Maybe this does not make much sense tough...

PR #478, opened 2026-04-14
Time aggregated losses

albangossard:
- Add support of dhdt and average ice velocity losses
- Fix a few minor bugs in the MB VJP computation

Associated PRs:
- https://github.com/ODINN-SciML/Sleipnir.jl/pull/166
- https://github.com/ODINN-SciML/Huginn.jl/pull/148


docs/src/optimization.md
```diff
@@ -84,6 +88,15 @@ For this setting, the empirical error term can be defined as
...
 
+#### Time aggregated loss functions
+
+There are cost functions which cannot be written as $\int_{t\in\Tau}\int_{x\in\Omega} \ell(...)$ but which are rather of the form $\ell(\int_{t\in\Tau} ...)$.
+These functions need special treatment and they are defined as a subtype of `AggregatedLoss`.
+This class of losses include:
+
+  - `LossDhdt`: Loss function for the mean glacier surface elevation change computed over a given time window.
+  - `LossAvgV`: Loss function for the mean ice surface velocity computed over a given time window.
+
```
facusapienza21, line 99: I insist I don't think this is correct. The second is a special case of the former. As you mention 'over a given window', this sounds like you are making an average, which by definition is an integral. Then you can discretize such integral, but still continuous. In my logic, all is the continuous case as you can always have a "delta" in the integral and most things won't break. I understand you want to explain the difference between a pure discretized and pure continuus loss, but maybe I wouldn't called them AggregatedLoss (Actually aggregated make me think more of an integral than a discrete operator)

facusapienza21:
Hi @albangossard 

I am still a bit confured of the logic behind defining the aggregated loss. In my mind, there are two types of losses, discrete and continuous, and there both can be writen in a very similar way. The former will use a pre-specied grid with the snapshots of the data, and the second will use a quadrature for evaluating the integrant and then make a weighted sum. In both cases, the code should be very similar. The minimal changes in `gradient.jl` are ok and in a way are trying to address this, but I still found confussing the definition of aggregated loss: I feel there is a lot of repetition in this code, and if well some details are different than in the original losses, these can be unifying much better. 

I understand this does not make terrible changes as this is a new loss rather than a fundamental change in the API, so if you feel the need for merging this we can move forward. However, now the logic behind losses is a bit more obscure, and I would like in the future to unify this and have an umbrella for all the losses (which again, for me they are more similar than different, the only difference is where to evaluate things and how). 

Another question: why the name "aggregated" for a loss that is a difference between two points? Aggregate makes me think of sums, which makes me think of integrals... which is exactly the opposite of what you want to do here, right? 

Happy to talk more about this if you want, but I understand there is more need to have this changes merge. I understand if you don't want to re-define the complete logic behind this PR, but then maybe I will suggest oppening an Issue to fix this at some point in the future. What do you think?

src/inverse/SIA2D/gradient.jl
```diff
@@ -251,12 +283,19 @@ function SIA2D_grad_batch!(θ, simulation::Inversion)
                 H_itp = interpolate((t,), H, Gridded(Linear()))
                 H_ref_itp = useThickness ?
                             interpolate((tH_ref,), H_ref, Gridded(Linear())) : nothing
+                # When there is only one reference velocity data we use a constant interpolator
                 Vabs_ref_itp = useVelocity ?
-                               interpolate((tV_ref,), Vabs_ref, Gridded(Linear())) : nothing
+                               (length(tV_ref)>1 ?
+                                interpolate((tV_ref,), Vabs_ref, Gridded(Linear())) :
+                                t -> only(Vabs_ref)) : nothing
```
facusapienza21, line 290: Is this expected to happen often? If not, maybe add a warnign

src/inverse/SIA2D/gradient.jl
```diff
@@ -330,14 +400,11 @@ function SIA2D_grad_batch!(θ, simulation::Inversion)
...
             cb_adjoint_MB = if simulation.parameters.simulation.use_MB
-                tstopsMB = - Huginn.define_callback_steps(tspan, simulation.parameters.simulation.step_MB)
-                function stop_condition_MB(λ, t, integrator)
-                    Sleipnir.stop_condition_tstops(λ, t, integrator, tstopsMB)
-                end
                 # For the moment the time stepping used in the loss, and the one for the MB gradient computation must match
                 # The plan in the future is to be able to customize the time stepping for the MB gradient computation
                 # Cf https://github.com/ODINN-SciML/ODINN.jl/issues/373
-                DiscreteCallback(stop_condition_MB, effect_MB!)
+                PeriodicCallback(effect_MB!, simulation.parameters.simulation.step_MB;
```
facusapienza21, line 424: Was this an error from a previous commit?

albangossard:
Hi @facusapienza21
Thanks for taking the time to go through this PR!

First I would like to clarify something:
> In my mind, there are two types of losses, discrete and continuous, and there both can be writen in a very similar way.

I partially agree with this. Given how `gradient.jl` works, there is a type of loss that doesn't fall into these two categories, for example (given that $\ell$ is non linear):

$$\int_{x\in\Omega}\ell\left(\int_{t\in\tau} \hat V(t,x) d\mu_t(t), V(x) \right)d\mu_x(x)$$

I agree that we could compute this integral using a quadrature but at the same time `gradient.jl` heavily relies on the fact that loss functions are evaluated point-wise in time, and which is something we need to make the adjoint computation efficient.
This is exactly why I implemented these "aggregated" losses which are different than the classical losses and work with the snapshots of ice thickness required for the computations.

> I would like in the future to unify this and have an umbrella for all the losses

If you come up with an idea to unify this I would be more than happy to exchange/implement this!

Because EGU is coming and I'll be focused on other tasks, if we don't unify this quickly I would suggest merging this PR and opening an issue to keep track of this as you mentioned.

> Another question: why the name "aggregated" for a loss that is a difference between two points?

Because the loss depends on some aggregated results, but I agree that the name is misleading since it doesn't tell if it's the input or the output that is aggregated. What do you think of `TemporallyNonLocalLoss` ?
Actually the dhdt loss can be seen as an integral, it's just that $\mu_t = \delta_{t_1} - \delta_{t_0}$ in the time integral.

I hope this answers your questions, otherwise please let me know!

docs/src/optimization.md
```diff
@@ -84,6 +88,15 @@ For this setting, the empirical error term can be defined as
 
 with $\hat H(t_j,x_j)$ the predicted ice thickness at time $t_j$ and on the node of the simulation grid $x_j$.
 
+#### Time aggregated loss functions
+
+There are cost functions which cannot be written as $\int_{t\in\Tau}\int_{x\in\Omega} \ell(...)$ but which are rather of the form $\ell(\int_{t\in\Tau} ...)$.
+These functions need special treatment and they are defined as a subtype of `AggregatedLoss`.
```
JordiBolibar: Even if it's more verbose, maybe `TimeAggregatedLoss` would be more self-explanatory?

src/inverse/SIA2D/adjoint.jl
```diff
@@ -49,7 +49,7 @@ function VJP_λ_∂SIA∂H_discrete(
     Δy = glacier.Δy
 
     # First, enforce values to be positive
-    map!(x -> ifelse(x > 0.0, x, 0.0), H, H)
+    H = map(x -> ifelse(x > 0.0, x, 0.0), H)
```
JordiBolibar, line 52: Why do you want to do this out-of-place now? Is this part of the bug you found?

src/inverse/SIA2D/gradient.jl
```diff
@@ -251,12 +283,19 @@ function SIA2D_grad_batch!(θ, simulation::Inversion)
                 H_itp = interpolate((t,), H, Gridded(Linear()))
                 H_ref_itp = useThickness ?
                             interpolate((tH_ref,), H_ref, Gridded(Linear())) : nothing
+                # When there is only one reference velocity data we use a constant interpolator
```
JordiBolibar, line 286: Do you mean when there are more than one reference velocity?

src/inverse/SIA2D/VJPs.jl
```diff
@@ -104,19 +105,37 @@ function VJP_λ_∂MB∂H(VJPMode::EnzymeVJP, λ, H, simulation::Simulation, gla
 end
 
 function VJP_λ_∂MB∂H(VJPMode::DiscreteVJP, λ, H, simulation::Simulation, glacier, t)
```
JordiBolibar, line 107: How would this be compatible with other MB models? It seems hardcoded for TI models.

src/inverse/SIA2D/VJPs.jl
```diff
@@ -104,19 +105,37 @@ function VJP_λ_∂MB∂H(VJPMode::EnzymeVJP, λ, H, simulation::Simulation, gla
...
+
+        cache.iceflow.MB .= compute_MB(model.mass_balance, glacier.climate.climate_2D_step, step_MB)
+        MB = cache.iceflow.MB
+        MB_mask = cache.iceflow.MB_mask
+
+        # MB, MB_mask, MB_total = ifm.MB, ifm.MB_mask, ifm.MB_total
+        MB_mask .= ((H .> 0.0) .&& (MB .< 0.0)) .|| ((H .> 10.0) .&& (MB .>= 0.0))
+        mask_ice_disappear = (H[MB_mask] + MB[MB_mask]) .< 0.0 # Mask of where the ice will disappear after MB application
 
         # The snow term doesn't depend on the ice thickness, hence it is null
```
JordiBolibar, line 132: What snow term? Is this the case for TI models with 2 DDFs? Here I only see you using one.

src/losses/TimeAggregatedLosses.jl
```diff
@@ -0,0 +1,392 @@
+export LossDhdt, LossAvgV
+
+# Losses that depend on time aggregated quantities
+abstract type AggregatedLoss <: AbstractLoss end
+
+function loss(
```
JordiBolibar, line 7: Maybe add a comment saying that this is just a fallback method for any non declared instances of `AbstractLoss`? I found it confusing at the beginning.

src/losses/TimeAggregatedLosses.jl
```diff
@@ -0,0 +1,392 @@
...
+    tLoss = tLoss[begin:(end - 1)] # Discard last point t=t2
+    ind_pred = Sleipnir.indFromT(simulation.parameters.simulation.tspan, tLoss, t)
+    T = sum(dt)
+
+    # 2. Get the reference velocity
+    V_ref = only(simulation.glaciers[glacier_idx].velocityData.vabs)
+    Vx_ref = only(simulation.glaciers[glacier_idx].velocityData.vx)
+    Vy_ref = only(simulation.glaciers[glacier_idx].velocityData.vy)
+
+    # 3. Compute the predicted velocity Vx_pred, Vy_pred, V_pred
```
JordiBolibar: This comment doesn't match the code, right?

JordiBolibar: It's #4 actually

src/losses/AggregatedLosses.jl
```diff
@@ -0,0 +1,392 @@
...
+        end
+    end
+    @assert cnt==length(tLoss)
+
+    return ∂L∂H, ∂L∂θ
+end
+
+loss_uses_velocity(lossType::LossAvgV) = true
+
+function aggregated_loss(
```
JordiBolibar: Same comment as above.

JordiBolibar (commented):
Thanks for all this work! Some comments I have to better understand things. 

I still think it might be interesting to get @facusapienza21 to check this to catch some adjoint-related things where I'm likely to miss some details.

src/inverse/SIA2D/adjoint.jl
albangossard: Yeah this is a bug, since this changes `H` in the adjoint computation, we had the wrong `H` in the following MB VJP call. This fix improves the accuracy of the adjoint.

src/inverse/SIA2D/gradient.jl
albangossard: No in the lines below we use a constant interpolator when there is only one data point in time. Otherwise we build a piecewise linear interpolation. In order to build a piecewise linear interpolator you need at least two time steps.

src/inverse/SIA2D/VJPs.jl
albangossard: Yes it is hardcoded for a specific TI model. We should use multiple dispatch here. I created an issue for this: https://github.com/ODINN-SciML/ODINN.jl/issues/480

src/inverse/SIA2D/VJPs.jl
albangossard: You see only one because the accumulation term doesn't depend on `H`, so the derivative of MB with respect to `H` yields only one term. But in `Muninn/src/models/mass_balance/mass_balance_utils.jl` there are actually two terms:
```julia
return ((mb_model.acc_factor .* climate_2D_period.snow) .-
            (mb_model.DDF .* climate_2D_period.PDD)) ./ (step / (1/12))
```

docs/src/optimization.md
albangossard: Thanks for the suggestion, yes this is more self-explanatory and I'll update this in the code

src/losses/TimeAggregatedLosses.jl
albangossard: Done in c384377

docs/src/optimization.md
albangossard: c384377

src/losses/TimeAggregatedLosses.jl
albangossard: Comment wasn't at the right place, thanks
c384377

src/losses/AggregatedLosses.jl
albangossard: c384377

albangossard (commented):
Thanks for the feedback @JordiBolibar !

I agree it's worth double checking this, so I'm merging the other PRs and we keep this one open until we are sure of the implementation.
@facusapienza21 please let me know what you think of my previous answer and the new name suggested by Jordi: `TimeAggregatedLoss` Once I have your feedback I'll merge this PR next week

docs/src/optimization.md
```diff
@@ -84,6 +88,15 @@ For this setting, the empirical error term can be defined as
 
 with $\hat H(t_j,x_j)$ the predicted ice thickness at time $t_j$ and on the node of the simulation grid $x_j$.
 
+#### Time aggregated loss functions
+
+There are cost functions which cannot be written as $\int_{t\in\Tau}\int_{x\in\Omega} \ell(...)$ but which are rather of the form $\ell(\int_{t\in\Tau} ...)$.
```
facusapienza21, line 93: Can we here add more description? Probably the formula in the PR and mention in which context we may need these losses.

src/inverse/SIA2D/gradient.jl
facusapienza21: Just checking this again

src/inverse/SIA2D/gradient.jl
```diff
@@ -315,12 +377,20 @@ function SIA2D_grad_batch!(θ, simulation::Inversion)
                         prod(N) * normalization,
                         Δtj
                     )
+                    if length(tstopsAggregatedLoss)>0 &&
+                       any(isapprox.(t, tstopsAggregatedLoss))
+                        ind = Sleipnir.indFromT(
+                            simulation.parameters.simulation.tspan, [t], tstopsAggregatedLoss)[1]
+                        ∂ℓ∂H .+= ∂L∂H_aggregated_loss[ind]
+                    end
```
facusapienza21: Is there a risk here in taking the same contribution more than once due to the `isapprox`?

src/losses/TimeAggregatedLosses.jl
```diff
@@ -0,0 +1,375 @@
...
+    t1 = Sleipnir.datetime_to_floatyear(only(simulation.glaciers[glacier_idx].velocityData.date1))
+    t2 = Sleipnir.datetime_to_floatyear(only(simulation.glaciers[glacier_idx].velocityData.date2))
+
+    # Get the reference velocity
+    V_ref = only(simulation.glaciers[glacier_idx].velocityData.vabs)
+    Vx_ref = only(simulation.glaciers[glacier_idx].velocityData.vx)
+    Vy_ref = only(simulation.glaciers[glacier_idx].velocityData.vy)
+
+    avg_Vx_pred, avg_Vy_pred,
+    avg_V_pred = Huginn.averageV(θ, simulation, (t1, t2), lossType.step, t, H_pred)
```
facusapienza21, line 168: Nice! In the future I would suggest to have all integrals (including an average) been computed with a quadrature!!! Even if an homogeneous one is desired.

src/inverse/SIA2D/gradient.jl
albangossard: This happens every time we use a loss function that depends only on one point in time and which requires to compute the ice velocity. For example this is the case as long as we use `LossAvgV` since this loss requires to have only one ice velocity snapshot in `glacier.velocityData`.

src/inverse/SIA2D/gradient.jl
albangossard: Yes this was an error I fixed. It's just that in the forward we are using `PeriodicCallback` and not `DiscreteCallback`, and we had an inconsistency between the forward and the gradient computations for the last MB call. We didn't take this into account in the reverse, leading to a wrong gradient and the impossibility to retrieve the scalar A properly (synthetic experiment without noise) as long as MB was activated.

docs/src/optimization.md
albangossard: Is your comment still up to date given our recent discussion on slack?

I updated the documentation of `LossDhdt` to make it more explicit that this is a difference of quantities.

src/losses/TimeAggregatedLosses.jl
albangossard: Yes I agree this should be implemented with a quadrature! Since for the moment this feature is mainly implemented for Lisa, and since we will handle average velocities computed over a few years, the computational bottleneck is rather limited. I created an issue to keep track and implement this in the future https://github.com/ODINN-SciML/ODINN.jl/issues/484

docs/src/optimization.md
albangossard: Good idea, I updated the documentation

src/inverse/SIA2D/gradient.jl
albangossard: Well that's a good point. In theory it could, but in practice I don't think this happens because the tolerance is so small that the solver makes time steps multiple orders of magnitude larger than the tolerance.

albangossard: We should have a way to better trigger that if statement. If you think this is critical I can implement something based on a callback instead.

albangossard: Implemented in bce2f04!

PR #501, opened 2026-06-02
SciMLSensitivity automatic adjoints

JordiBolibar:
This pull request introduces support for classical inversion of the sliding coefficient `C` (in addition to the existing support for `A`) in the ODINN framework, updates the law parameterizations to be compatible with SciMLSensitivity adjoints, and adds comprehensive documentation for configuring gradient computation. The changes include a new inversion script for `C`, enhancements to the `GriddedInv` component to handle both `A` and `C` with proper bounds and initialization, and the addition of a new `LawC` for differentiable sliding coefficient inversion.

The differentiability of `MultiLoss` and the regularizations is still WIP. There are still problems with Enzyme.jl that should be addressed, but at least this should provide the possibility of doing classical inversions of `C` without regularization. 

**Key changes:**

### New features and scripts

* Added `scripts/classical_inversion_C.jl`, a new example script for performing classical gridded inversion of the sliding coefficient `C` using the new `LawC` and `GriddedInv(:C)` components.
* Enhanced `scripts/classical_inversion.jl` to use the improved gradient configuration and law parameterization.

### Law parameterization and compatibility

* Added `LawC` to `src/laws/Laws.jl`, enabling differentiable inversion of the sliding coefficient `C` with tanh-based parameterization and proper handling for both scalar and gridded cases. The law is compatible with both manual and SciMLSensitivity adjoints. [[1]](diffhunk://#diff-044b2342534085566423305e087419967a0faaa1233f1e7b27ec278b01db6d18L1-R1) [[2]](diffhunk://#diff-044b2342534085566423305e087419967a0faaa1233f1e7b27ec278b01db6d18L455-R551)
* Updated `LawA` to determine `callback_freq` based on the selected adjoint method, ensuring correct gradient flow for both manual and SciMLSensitivity adjoints. [[1]](diffhunk://#diff-044b2342534085566423305e087419967a0faaa1233f1e7b27ec278b01db6d18R405-R421) [[2]](diffhunk://#diff-044b2342534085566423305e087419967a0faaa1233f1e7b27ec278b01db6d18L427-R447)

### Model and inversion component improvements

* Extended `GriddedInv` to support both `:A` and `:C` inversion, with min/max bounds determined by the parameter and robust initialization even if glacier values are out of bounds. [[1]](diffhunk://#diff-8d1a271fd82ea1a0a69f5f736754fd92fc8bd6f419e59c559e2ffc3fffb35958L43-R45) [[2]](diffhunk://#diff-8d1a271fd82ea1a0a69f5f736754fd92fc8bd6f419e59c559e2ffc3fffb35958L52-R79)

### Adjoint method and API updates

* Added and exported `SciMLSensitivityAdjoint` as a new adjoint method, making it available for use in parameter configuration.

### Documentation

* Added a comprehensive section to `docs/src/sensitivity.md` detailing how to configure gradient computation via `UDEparameters`, including explanations of the new adjoint method options, `sensealg`, and `optim_autoAD` fields, along with practical configuration examples.


src/laws/Laws.jl
```diff
@@ -452,13 +467,83 @@ function LawA(params::Sleipnir.Parameters; scalar::Bool = true)
...
+    end
+    # No p_VJP! here: LawC is designed for SciMLSensitivityAdjoint, which replays the RHS
+    # during the backward pass and therefore never needs an explicit analytical VJP for C.
+    # Manual adjoint methods (ContinuousAdjoint, DiscreteAdjoint) are not yet supported
+    # for C inversion; using them will yield a zero gradient for θ.C.
+    f! = let min_C = min_C, max_C = max_C
+        function (cache, inp, θ)
+            val = @. min_C +
+                     (max_C - min_C) * (tanh.(θ.C[Symbol("$(cache.glacier_id)")]) + 1) /
+                     2
```
albangossard: For the basal sliding coefficient I don't think that `tanh` as it is implemented right now is an appropriate parameterization for two reasons:
1. We want to allow null basal sliding values.
2. An initialization with a given C value (whatever scalar or matrix) in Glacier2D should result in the same values with the parameterization for the initialization. Currently since you cannot invert zero values, you are modifying the initialization of the parameterization (cf `GriddedInv`).

So I don't think we should have a `minC` parameter, and this is not clear for me that `maxC` is also needed (maybe for numerical stabilization?). Depending on the strategy I would recommend to use:
- **Without `maxC`** The negative log-sigmoid `y=log(1/(1+exp(x)))` which has the advantage of being differentiable everywhere and smooth (important as soon as you have no spatial regularization over C) + it vanishes quickly towards zero which allows having not truly C=0 but something very close.
- **With `maxC`** A `tanh` parameterization with `min_C=0` hard-coded in the code above.

In order to have a consistent initialization for values at the limit of the range of your parameterization (let's say those indices are in `mask`), I would recommend to initialize with values such that you are very close. For example with `tanh`, `min_C=0` and `max_C=1` you could simply set `θ[mask] .= -5` which would yield `C=5e-5` , this is already very close to the initialization.

src/models/trainable_components/GriddedInv.jl
```diff
@@ -49,10 +51,32 @@ mutable struct GriddedInv{
...
+            minv = isnothing(minval) ? params.physical.minC : minval
+            maxv = isnothing(maxval) ? params.physical.maxC : maxval
+            minv < maxv ||
+                error("GriddedInv: expected minC < maxC, got minC=$(minv), maxC=$(maxv)")
+            # If any glacier.C value is outside the open interval (minv, maxv), seed from
+            # the midpoint. This mirrors GriddedInv(:A) where glacier.A is always in-bounds
+            # by construction; glacier.C defaults to 0 (no sliding), which is physically
+            # correct but out-of-range for the atanh mapping.
+            if any(x -> x <= minv || x >= maxv, collect(θ))
+                midv = Sleipnir.Float((minv + maxv) / 2)
```
albangossard: `Glacier2D` exposes an API to initialize `A` and `C` for example and we should initialize the parameterization to be as close as possible to these user provided values.

test/runtests.jl
```diff
@@ -144,6 +144,8 @@ ENV["GKSwstype"] = "nul"
                 thres = [5e-4, 7e-7, 2e-3])
             @testset "SciMLSensitivity adjoint with Enzyme VJP vs finite differences" test_grad_finite_diff(
                 ODINN.SciMLSensitivityAdjoint(); thres = [1e-5, 1e-7, 1e-5])
+            @testset "SciMLSensitivity auto-adjoint vs manual ContinuousAdjoint for LawA" test_grad_sciml_vs_manual(thres = [
+                1e-3, 1e-13, 1e-3])
```
albangossard, line 148: Not that bad, that's great!

docs/src/sensitivity.md
```diff
@@ -98,3 +98,87 @@ Here, we compile the main considerations and things that need to be taken into a
...
+# Quadrature adjoint — similar to GaussAdjoint but uses a dedicated
+# quadrature library for the time integral. Can be more accurate for
+# smooth solutions but requires more memory than the interpolating variant.
+sensealg = QuadratureAdjoint(autojacvec = SciMLSensitivity.EnzymeVJP())
+
+# Backsolve adjoint — reconstructs the forward solution on the fly during
+# the backward pass instead of storing it. Lowest memory footprint but
+# numerically unstable for the SIA; not recommended in practice.
+sensealg = BacksolveAdjoint(autojacvec = SciMLSensitivity.ZygoteVJP())
+```
```
albangossard, line 144: Have you tried to run these adjoints? Because when I tried a few months back only the `InterpolatingAdjoint` was working for us. That was because over these three adjoints one wasn't supporting SciMLStructs, and the two others were unstable with our adjoint ODE.
If this is still the case that's worth mentioning it in the documentation.

Note that with `BacksolveAdjoint` you mentioned `ZygoteVJP` but that doesn't work with our code as everything in the SIA is in-place.

docs/src/sensitivity.md
```diff
@@ -98,3 +98,87 @@ Here, we compile the main considerations and things that need to be taken into a
...
+
+### Layer 3 — `optim_autoAD`: outer AD signal
+
+`optim_autoAD` acts as a required signal that tells ODINN which outer differentiation approach is being used. Importantly, ODINN always builds the `OptimizationFunction` internally with `NoAD()` and a custom `grad!` function — `optim_autoAD` is never passed directly to the optimizer machinery. Its role is to validate that the configuration is self-consistent:
+
+  - `Optimization.AutoZygote()` **must** be used with `SciMLSensitivityAdjoint()`. ODINN enforces this with an assertion at runtime. Internally, the custom `grad!` calls `Zygote.gradient` on the loss, which differentiates through the `solve(...)` call; SciMLSensitivity intercepts the ODE adjoint via the `sensealg`.
```
albangossard: To explain why we are providing the grad function ourselves even with SciMLSensitivity
```suggestion
  - `Optimization.AutoZygote()` **must** be used with `SciMLSensitivityAdjoint()`. ODINN enforces this with an assertion at runtime. Internally, the custom `grad!` calls `Zygote.gradient` on the loss, which differentiates through the `solve(...)` call; SciMLSensitivity intercepts the ODE adjoint via the `sensealg`. This allows us to handle the differentiation within each process with multiprocessing properly.
```

docs/src/sensitivity.md
```diff
@@ -98,3 +98,87 @@ Here, we compile the main considerations and things that need to be taken into a
...
+  - `ODINN.NoAD()` must be used with the manual adjoint methods (`ContinuousAdjoint`, `DiscreteAdjoint`). In this case the custom `grad!` calls `SIA2D_grad!` directly, which runs the hand-written adjoint without any outer AD.
```
albangossard: It's not really that it **must** be used, it's just that the parameter has no impact, I don't know if you want to mention this in the documentation.

docs/src/sensitivity.md
```diff
@@ -98,3 +98,87 @@ Here, we compile the main considerations and things that need to be taken into a
...
+
+### Inner VJP for the manual adjoints
+
+When using `ContinuousAdjoint` or `DiscreteAdjoint`, the `VJP_method` field inside those structs controls how the Jacobian of each law is evaluated during the backward pass. The available options are `AbstractVJPMethod` subtypes:
+
+  - `DiscreteVJP(regressorADBackend = DI.AutoMooncake())` (default): Manual pullback implementation, with an AD backend for the regressor VJPs.
```
albangossard: That backend isn't always used in practice.
```suggestion
  - `DiscreteVJP(regressorADBackend = DI.AutoMooncake())` (default): Manual pullback implementation, with an AD backend for the regressor VJPs when the user doesn't provide how to compute them.
```

docs/src/sensitivity.md
```diff
@@ -98,3 +98,87 @@ Here, we compile the main considerations and things that need to be taken into a
...
+  - `EnzymeVJP()`: Calls Enzyme directly to evaluate the VJP, without going through DifferentiationInterface.
```
albangossard: That `DifferentiationInterface` thing is quite different than these `DiscreteVJP`/`EnzymeVJP`/`ContinuousVJP`. The real thing with `EnzymeVJP` is that it differentiates the entire `SIA2D` and it's not really a question of using `DifferentiationInterface` or not, but rather treating the VJPs of the regressors with a special care (could be `DifferentiationInterface` or user provided VJPs).

docs/src/sensitivity.md
```diff
@@ -98,3 +98,87 @@ Here, we compile the main considerations and things that need to be taken into a
...
+)
+```
+
+**Using ODINN's continuous adjoint with Enzyme VJPs:**
+
+```julia
+# optim_autoAD = NoAD() because SIA2D_grad! provides the gradient directly.
+UDE = UDEparameters(
+    grad = ContinuousAdjoint(VJP_method = EnzymeVJP()),
+    optim_autoAD = ODINN.NoAD()
```
albangossard: You don't really need that parameter

albangossard (commented):
Thanks for this work @JordiBolibar! I'm happy to see that the SciML gradients work with custom regressors thanks to your fix. Your additions to the documentation are also great!
Please find some comments below just by looking at the code. I'll perform a second review once I start playing with it.

src/laws/Laws.jl
JordiBolibar: Thanks for the feeback! I think it would be good to keep `maxC`, since as you mentioned, a too large C value could produce numerical instabilities in the solver. I'll go for the second option.

docs/src/sensitivity.md
JordiBolibar: No, I haven't tried them. I suggest we open an issue to investigate that. I have modified the `ZygoteVJP` part, and replaced it by `EnzymeVJP`.

JordiBolibar: I opened #502 to track this.

albangossard (approved):

PR #509, opened 2026-06-09
Fix MultiLoss differentiation with Zygote

albangossard:
The main purpose of this PR is to fix the `MultiLoss` differentiation with `Zygote`.
I first suspected the non concrete types to be responsible of the issue. It turns out this is not the case and the issue was due to a `map` messing up with `Zygote`. Since it's always a good practice to have concrete types I suggest that we keep the changes I made.
There are utility functions in Sleipnir to check that all fields of a nested struct are of concrete type, cf https://github.com/ODINN-SciML/Sleipnir.jl/pull/184
Similar changes were made in Huginn, cf https://github.com/ODINN-SciML/Huginn.jl/pull/165

ODINN does not need the associated PRs to run, but the tests need them since we rely on the utilities mentioned above.


docs/src/functional_inversion.jl
```diff
@@ -236,5 +236,5 @@ plot_law(functional_inversion.model.iceflow.A, functional_inversion, law_input,
 # Since we have a very limited range of temperature coverage with these 4 glaciers, we have only captured a small portion of the ground truth law, which has an almost linear form. If we wanted to learn the full dynamics of the Cuffey and Paterson synthetic law, we would need to have glaciers that cover a wider range of temperatures.
 
 # Just to have some additional context, here is how the synthetic law looks like for the full range of temperatures:
-plot_law(prediction.model.iceflow.A, functional_inversion,
+plot_law(prediction.model.iceflow.A, prediction,
```
JordiBolibar, line 239: Why this change? The file is `functional_inversion.jl`.

JordiBolibar (approved):
Just a very small comment. Thanks for taking care of this!

docs/src/functional_inversion.jl
albangossard: At this line you want to plot the synthetic law, so why would you need `functional_inversion`?
I had to change this because now we cannot change the type of an instantiated object at runtime, meaning that the line `simulation.model.trainable_components.θ = θ` (`eval_law` in `ODINN/src/laws/laws_utils.jl` called in `plot_law_1d`) should change only the values and not the type which is what used to happen.

JordiBolibar:
It seems there's still an error in the tests:

```julia
> Testing target A with SciMLSensitivityAdjoint() and LossH
[ Info: Running forward in-place PDE ice flow model
Processing glacier RGI60-11.03638 for PDE forward simulation
┌ Warning: Using light mode of neural network
└ @ ODINN ~/work/ODINN.jl/ODINN.jl/src/models/trainable_components/ML_utils.jl:25
SciMLSensitivity adjoint with Enzyme VJP vs finite differences: Error During Test at /home/runner/work/ODINN.jl/ODINN.jl/test/runtests.jl:151
```

albangossard:
> It seems there's still an error in the tests:

I'm currently debugging this, it seems that the Enzyme segfault with SciMLSensitivity is back

albangossard:
Since the full tests are passing and since I cannot reproduce the segfault locally and it happens only with the fast tests within the CI, I'm going to merge this PR.

PR #511, opened 2026-06-18
Sync main from dev

albangossard:


albangossard:
@JordiBolibar I fixed the full tests CI by splitting them into different groups. This still runs in test mode (in opposition to the fast tests which use `include`).
I had to deactivate the `LossAvgV` test because it fails in the CI and this happens only with macOS. We checked with Lisa and that test passes on her laptop. Since it is already tested with Ubuntu I don't think this is critical.
Let me know what you think of that workaround!

JordiBolibar:
Sounds good! Thanks for all this work. Let's keep working like this for now.

PR #512, opened 2026-06-18
speed-up custom NN test

albangossard:
@facusapienza21 with recent versions of the packages we are using, the custom NN with pure diffusivity test no longer passes within the CI. This is because of an out of memory error. Don't you mind if I deactivate that test?

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -202,15 +202,16 @@ ENV["GKSwstype"] = "nul"
                 thres = [5e-3, 1e-6, 5e-3], target = :D, loss = LossV())
         end
     end
-    if (GROUP == "All" && (!CI || !Sys.isapple())) || GROUP == "Core7"
-        # Skip this test on macOS when running the "Full tests" CI because it is too slow and produces a timeout error (>6h)
-        @testset "Adjoint method of SIA equation with pure D as target and custom NN" begin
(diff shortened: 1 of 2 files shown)


JordiBolibar:
Can we find a fast and easy way to make it more lightweight? Maybe by coarsening the grid?

albangossard:
> Can we find a fast and easy way to make it more lightweight? Maybe by coarsening the grid?

We could but I don't think this brings anything since what we want to test with target D is the quality of the interpolation. A coarser spatial grid would make the interpolation better since we would have the same amount of points for a reduced range of values.

In my opinion this test doesn't bring anything we already have except the differentiation with respect to θ of the neural network part. This step relies on DifferentiationInterface and we can trust the result, meaning that if this works for a small network (the other tests), this is going to work for larger ones.

facusapienza21:
Auch, I would also prefer not do deactivate this test... Do you think this is covered by the other tests? In my experience, it was quite hard to make this example to work, reason why it would be good to keep it. What about reducing even more the size of the glacier we use to test??? Keep the interpolation low for memory issues, but also reduce the size of the glacier??? We can also just have continuiuty test to test this code "runs" with very low interpolation but doesn't give the right gradient. What do you think?

albangossard:
> Do you think this is covered by the other tests?

The custom neural network is not covered by the other tests, but all the other ingredients are, so that's why I don't think this test brings much.

Still, I tried to keep that test in the CI. What I did is that I reduced the glacier grid (downscaling factor of 8 instead of 4), I reduced the number of interpolation points (50 instead of 200), and I subsampled θ to test only 25 components. The test now takes 1 minute on my laptop.
@facusapienza21 please let me know what you think of that solution and if that suits you I'll merge this PR

PR #518, opened 2026-06-25
Remove parametric type for initial_conditions in TrainingStats

### LINKED ISSUE #517: Mismatch between `tstopsAggregatedLoss` and `tLoss`
When trying to inverse real data with `LossAvgV`, at some point there is a mismatch in `backward_time_aggregated_loss` between the tstops available in the initial steps vs in the Loss which led that at this step: `if t[i] in tLoss` ; some `t[i]` are not in `tLoss` and the condition `cnt==length(tLoss)` is not matched.

albangossard:
- Remove parametric type for `initial_conditions` in `TrainingStats` as we need flexibility to change the type at runtime

Fixes #517

### DIFF
--- a/src/inverse/SIA2D/gradient.jl
+++ b/src/inverse/SIA2D/gradient.jl
@@ -102,7 +102,7 @@ function SIA2D_grad_batch!(θ, simulation::Inversion)
             tstopsDiscreteLoss, tstopsAggregatedLoss)))
 
         @assert length(t) == length(tstops) "The size of tstops does not match with the size of the reference times."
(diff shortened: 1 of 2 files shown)


JordiBolibar (approved):
This doesn't introduce any type instabilities?

albangossard:
> This doesn't introduce any type instabilities?

Yes this does but unfortunately we have no choice. Fixing these type instabilities would require to redesign a large portion of the inversion workflow by re-instantiating `inversion` and its fields. Note that we already had type instabilities with respect to `θ`

JordiBolibar:
@albangossard Fair enough, feel free to merge whenever you want.

PR #519, opened 2026-06-25
Fix cache dimensions bug

### LINKED ISSUE #516: `cache.vjp_θ` takes `A` and `IC` for and inversion of `A` only
While trying to invert `A` using a `GriddedInv` with `InitialCondition`, an error is raised in `Laws.jl` when executing this line:

`cache.vjp_θ .= vec(ret.A[Symbol("$(cache.glacier_id)")])`

This happens because `cache.vjp_θ `contains both `A` and `IC`, not just `A`, which causes a dimension mismatch.

albangossard:
- Fix cache dimension bug for joint inversions #516 
- Make `t₀` mandatory when declaring `InitialThicknessRegularization`. The default value used to be 1994.0 and when `t₀` was outside of `tspan`, this led to obscure errors. Since we have no way to check the value of `t₀` in `InitialThicknessRegularization`, we cannot add an assert. The user now has to provide an explicit value.
- Add a joint inversion test (gridded A + IC)

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -247,7 +247,7 @@ ENV["GKSwstype"] = "nul"
                 ContinuousAdjoint(VJP_method = DiscreteVJP()); thres = [1e-4, 1e-8, 1e-4],
                 functional_inv = false, scalar = false,
                 loss = MultiLoss(
-                    losses = (LossH(), InitialThicknessRegularization(t₀ = 2010.0)), λs = (
+                    losses = (LossH(), InitialThicknessRegularization(2010.0)), λs = (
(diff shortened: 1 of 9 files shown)


test/runtests.jl
```diff
@@ -246,6 +242,15 @@ ENV["GKSwstype"] = "nul"
                     functional_inv = false, scalar = true, loss = LossAvgV(), aggregated_loss = :avgV)
             end
         end
+        @testset "Joint inversion" begin
```
JordiBolibar, line 245: Great! Thanks!

JordiBolibar (approved):

PR #510, opened 2026-06-15
TIM calibration using Hugonnet geodetic data

JordiBolibar:
This pull request introduces improvements to how mass balance (MB) models are selected and managed, especially for workflows where MB models can be calibrated per glacier. The main changes ensure that the correct MB model is used for each glacier throughout the simulation and inversion processes. This involves updating function signatures, extracting the MB model per glacier where needed, and optionally calibrating MB models per glacier during inversion setup.

**Mass Balance Model Selection and Calibration:**

* The MB model is now extracted per glacier using `get_mb_model` in several places, ensuring that each glacier uses its correct (possibly calibrated) MB model. This change affects simulation batching, VJP computations, and callback construction. [[1]](diffhunk://#diff-8bec99d33808bba78380aad81f8d98f84a6b66ab144124ceb3059e8b90756bedL89-R91) [[2]](diffhunk://#diff-6f5e6086914bdd68d70f2c4cbf55f50cec34e03f89df791f049e7275dc5dfa2bR89-R92) [[3]](diffhunk://#diff-6f5e6086914bdd68d70f2c4cbf55f50cec34e03f89df791f049e7275dc5dfa2bL114-R119) [[4]](diffhunk://#diff-c2bff6412ed5a8707275648f5474f356c709bdf20878069c660208d2f076754dL504-R510)
* The inversion setup (`Inversion`) now optionally calibrates the MB model per glacier if the configuration specifies it, enabling per-glacier calibration workflows.

**Function and API Updates:**

* The `MB_wrapper!` function and related VJP functions have updated signatures to accept an explicit MB model argument, reflecting the new per-glacier model selection. [[1]](diffhunk://#diff-6f5e6086914bdd68d70f2c4cbf55f50cec34e03f89df791f049e7275dc5dfa2bL71-R78) [[2]](diffhunk://#diff-6f5e6086914bdd68d70f2c4cbf55f50cec34e03f89df791f049e7275dc5dfa2bR105)

These changes together improve the flexibility and correctness of simulations and inversions involving mass balance models, especially in scenarios where MB models are glacier-specific.

PR #521, opened 2026-07-24
Move abstract losses definition to Sleipnir so that velocity product can be determined based on the loss type

albangossard:
PR associated to https://github.com/ODINN-SciML/Sleipnir.jl/pull/191


albangossard:
@JordiBolibar since I need the changes of #510 in order to run the tests, I'm merging this PR and we will make sure the CI passes directly in dev

PR #527, opened 2026-08-01
Fix velocity-loss chain rule and continuous-adjoint IC gradient

JordiBolibar:
Found while debugging a joint A+IC inversion with a velocity loss, where H0 stayed pinned to the Millan22 prior instead of being updated by the optimizer. Two separate bugs.

`LossV`/`LossAvgV`'s `:abs` component computed `d(loss)/d(Vx,Vy)` with a secant slope (`∂l/∂V · (Vx_pred - Vx_ref)/(V_pred - V_ref)`) instead of the chain rule through `V = √(Vx² + Vy²)` (`∂l/∂V · Vx/V`). The secant only matches the true gradient when predicted and reference flow directions coincide, which holds closely enough for SIA that no existing test caught it, but blows up as `V_pred → V_ref`. Fixed by composing with `VJP_λ_∂V∂Vxy` instead. Added `test_grad_V_from_Vxy`, a solver-free FD test against the real loss functions, to guard against regressing to the secant.

The continuous adjoint's λ0 extraction at the IC boundary dropped the t0 loss deposit: the callbacks do apply it during the reverse solve, but `sol_rev` interpolated at `-tspan[1]` does not carry that jump, which made `dL/dH0` too small. Fixed by applying `effect_loss!`/`effect_aggregated_loss!` to λ0 after the interpolation. Couldn't cover this with an automated FD test: perturbing H0 shifts the adaptive solver's step sequence and the resulting noise swamps the effect being measured. Checked by hand instead, via δt-refinement (the fixed adjoint's IC gradient is stable under solver step refinement, naive FD isn't) and by cross-checking against an FD-trustworthiness metric across several seeds.

Also added a docstring warning + runtime `@warn` on `DiscreteAdjoint`: its IC gradient is a real method limitation, not a bug, a single explicit-Euler reverse step is unstable/badly inaccurate for the stiff SIA at practical (monthly) time steps. Points to `ContinuousAdjoint` for IC inversion instead.


src/inverse/SIA2D/gradient.jl
```diff
@@ -505,7 +508,17 @@ function SIA2D_grad_batch!(θ, simulation::Inversion)
 
             # Compute gradient wrt initial condition because this is not taken into account in the quadrature
             if haskey(θ, :IC)
-                λ₀ = sol_rev(-tspan[1])
+                λ₀ = copy(sol_rev(-tspan[1]))
+                # The loss deposits at t₀ are not applied by the discrete callbacks at the
+                # reverse-integration boundary (same reason the t₁ contribution is added
+                # manually to λ₁ above). dL/dH₀ = λ(t₀⁻) must include the t₀ jump, so add
+                # it explicitly here; otherwise the initial-condition gradient is too small.
```
albangossard, line 515: This part took me a long time to understand. I did not get why you were applying `effect_loss!`/`effect_aggregated_loss!` again because this **is** already included in the reverse ODE solve for t=t₀. The issue mentioned earlier refers to `cb_adjoint_loss`/`cb_adjoint_aggregated_loss` not hitting t=t₁ but the reason you need to apply it again is because you use `sol_rev` which is interpolated at `-tspan[1]`. The sentence "same reason the t₁ contribution is added manually to λ₁ above" is misleading and I suggest either to clarify it or to remove it.

albangossard (approved):
Good catch fixing this @JordiBolibar!

I have one minor suggestion to clarify what the fix is about.

Also could you fix the CI in `dev` before merging this PR please? I'm afraid that we will introduce silent bugs since the CI does not pass.

albangossard:
One thing that came to my mind if we want to improve our test coverage is to check the gradient using only IC, without any target. With `losses = (LossAvgV(component = :abs),), λs = (1.0,)),` we would be able to catch such errors (actually this is what I've started with during debugging). This is not urgent, so feel free to implement this inside this PR, or I can give it a try in a few weeks :)

PR #529, opened 2026-08-28
feat: flexible gradient check

albangossard:


docs/src/sensitivity.md
```diff
@@ -180,3 +180,16 @@ UDE = UDEparameters(
     grad = ContinuousAdjoint(VJP_method = EnzymeVJP()),
 )
 ```
+
+## Numerical verification of the gradient
```
JordiBolibar, line 184: Very nice wrapper!

ext/FiniteDifferencesExt.jl
```diff
@@ -0,0 +1,175 @@
...
+        params = simulation.parameters
+        n_params = length(θ)
+        if n_params > max_params
+            # Evaluate gradient on subset of parameters to save some computation
+            @info "Testing gradient with a subset of parameters of size $(max_params) since the original parameter vector θ is of dimension $(n_params)."
+
+            # Component array with binary entry
+            θ_mask = θ .== nothing
+
+            for key in keys(θ)
```
JordiBolibar, line 117: Is this only for manual adjoints? Because for the automatic adjoints we should also cover C. If so, we should also mention this in the docstring.

JordiBolibar (approved):

ext/FiniteDifferencesExt.jl
albangossard: This works with any backend, so this should work if you try to check the gradient wrt C with SciMLSensitivity
Function updated in 1547e33

PR #530, opened 2026-08-28
Fix Core1/3/8/9 and Documentation CI failures on dev

JordiBolibar:
#510 merged without CI ever running, and Sleipnir 0.15's move to a per-day PDD/snow downscaling broke the MB-coupled adjoint machinery in several places. This fixes all of it.

Sleipnir 0.15 replaced the monthly-aggregate temperature-index downscaling with a per-day loop. The discrete MB VJP still linearized the old monthly aggregate and dropped the snow term entirely, even though snow now depends on ΔS (hence H) — this is Core3. Separately, `gradient.jl` linearized the SIA VJP at the post-MB state when the SIA flow over each interval actually ends at the pre-MB state; this explains why the Enzyme MB VJP variant also failed despite Enzyme differentiating the real code. Both are fixed, gated so they're a no-op when `use_MB = false` (zero effect on Core2/4/5/6/7/10).

The `dhdt` loss test (Core8) melted its glacier out under the corrected PDD accounting — the old code under-counted melt by clipping PDD after summing instead of per day. The forcing (`DDF = 15e-3` over 1980-2019) was never physical to begin with; the fix recalibrates it against real geodetic DDFs (3.9-8.8e-3 for these glaciers) rather than loosening tolerances.

Core9 turned out to be a separate bug: `Inversion`'s constructor discarded `calibrate_MB_model!`'s return value, so the optimizer trained against the uncalibrated MB model while `generate_ground_truth` (via `Prediction`) used the calibrated one. Using the return value fixes it. This is the same bug Muninn's `calibrate_MB_model!` → `calibrate_MB_model` rename ([Muninn#66](https://github.com/ODINN-SciML/Muninn.jl/pull/66)) hit three separate times across the ecosystem — the bang was misleading, since calibration changes `mass_balance`'s type and can never mutate in place. This branch migrates to the new name ([Huginn#173](https://github.com/ODINN-SciML/Huginn.jl/pull/173), [MassBalanceMachine#28](https://github.com/ODINN-SciML/MassBalanceMachine.jl/pull/28) are the sibling migrations).

Two more VJP bugs found while chasing this: both `MB_wrapper!` (Enzyme path) and the discrete MB VJP dropped `temp_bias` when calling `downscale_2D_climate!`, so any calibrated model with nonzero `temp_bias` got an adjoint of the wrong function; and the discrete VJP's `compute_MB` call used `model.mass_balance` instead of the already-selected per-glacier `mb_model`, throwing under calibration. Both fixed with regression tests.

`test_grad_finite_diff` and `inversion_test` now default to `calibrate_MB = false`, so each test's own `TImodel1` is what's actually exercised instead of being silently replaced by geodetic calibration — `calibrate_MB = false` reproduces the pre-#510 baseline loss (~2e-13) to 15 digits, confirming there's no residual gradient bug once this is restored. One `calibrate_MB = true` gradient test is added back for coverage of the per-glacier vector path.

Documentation was crashing before any of the above could even run, on a `Glacier2D` vector `show` hitting `eltype` on a `UnionAll` ([Sleipnir#196](https://github.com/ODINN-SciML/Sleipnir.jl/pull/196)). Fixing that exposed two tutorial bugs it had been masking: `smb_calibration.jl` had the same discarded-`calibrate_MB_model` bug, plus referenced a `glacier.geodetic_MB` field that never existed (the real one is `glacier.dhdtData.dhdt`); `results_plotting_tutorial.jl` imported both `ODINN` and `Sleipnir`, making `Model` ambiguous.

Depends on Sleipnir 0.15.2, Muninn 0.13.0, and Huginn 0.15.1 being released before merge — all three PRs above are ready.

Local verification (CI_FAST, all groups on ubuntu, `Pkg.test()`-equivalent): Core1 60/14/0, Core3 45/28/0, Core8 24/16/0, Core9 12/12/0 (pass/broken/fail), Documentation builds clean end to end.


test/test_utils.jl
```diff
@@ -76,9 +76,16 @@ Returns:
     to normalize and compute the relative error.
 """
 function stats_err_arrays(a::T, b::T) where {T}
-    ratio = sqrt(sum(a .^ 2)) / sqrt(sum(b .^ 2)) - 1
-    angle = sum(a .* b) / (sqrt(sum(a .^ 2)) * sqrt(sum(b .^ 2))) - 1
-    relerr = sqrt(sum((a - b) .^ 2)) / sqrt(sum((a) .^ 2))
+    norm_a = sqrt(sum(a .^ 2))
+    norm_b = sqrt(sum(b .^ 2))
+    # Both gradients are exactly zero (e.g. the tested parameter has no influence
+    # on the loss for this scenario): they trivially agree, avoid a 0/0 NaN.
```
albangossard: This could be a gradient or anything else
```suggestion
    # Both vectors are exactly zero: they are trivially aligned, avoid a 0/0 NaN.
```

src/inverse/SIA2D/VJPs.jl
```diff
@@ -74,7 +74,7 @@ function MB_wrapper!(MB, H, simulation, glacier, mb_model, step)
 
     # Below we call the functions that are inside MB_timestep! manually
     # This is because get_cumulative_climate! cannot be differentiated with Enzyme, so it is called beforehand in the VJP function to retrieve the cumulative climate
-    downscale_2D_climate!(glacier)
+    downscale_2D_climate!(glacier; temp_bias = get_temp_bias(mb_model))
```
albangossard, line 77: I would make `temp_bias` mandatory, otherwise this is super error prone

src/inverse/SIA2D/VJPs.jl
```diff
@@ -118,15 +117,30 @@ function VJP_λ_∂MB∂H(VJPMode::DiscreteVJP, λ, H, simulation::Simulation, g
...
+        ∂PDD∂S = zero(glacier.S)
+        ∂snow∂S = zero(glacier.S)
+        for d in eachindex(temp_vec)
+            T_d, g_d, p_d = temp_vec[d] + temp_bias, grad_vec[d], prcp_vec[d]
+            @. ∂PDD∂S += ifelse(T_d + g_d * ΔS > 0, g_d, 0.0)
+            @. ∂snow∂S += ifelse((2 - T_d - g_d * ΔS > 0) & (2 - T_d - g_d * ΔS < 2),
+                -p_d * g_d / 2, 0.0)
+        end
+        ∂MB∂S = (PRECIP_UNIT_CONVERSION * mb_model.prcp_fac .* ∂snow∂S .-
+                 mb_model.DDF .* ∂PDD∂S) ./ (step_MB / (1 / 12))
```
albangossard, line 141: Excellent!! Thank you for this fix!

src/inverse/SIA2D/gradient.jl
```diff
@@ -303,6 +307,26 @@ function SIA2D_grad_batch!(θ, simulation::Inversion)
                 throw("Interpolation method for continuous adjoint not defined.")
             end
 
+            # Interpolant used only for the SIA VJP linearization. The forward SIA flow over
```
albangossard:
```suggestion
            # Linear interpolant (as above) used only for the SIA VJP linearization. The forward SIA flow over
```

src/inverse/SIA2D/gradient.jl
```diff
@@ -318,7 +342,7 @@ function SIA2D_grad_batch!(θ, simulation::Inversion)
                     t = -τ
```
albangossard, line 342: You need to replace `H_itp` by `H_itp_SIA` in the `let` above

src/inverse/SIA2D/gradient.jl
```diff
@@ -303,6 +307,26 @@ function SIA2D_grad_batch!(θ, simulation::Inversion)
                 throw("Interpolation method for continuous adjoint not defined.")
             end
 
+            # Interpolant used only for the SIA VJP linearization. The forward SIA flow over
+            # each interval ends at the pre-MB state (MB is applied at the right endpoint),
+            # but H stores post-MB states. Interpolate post-MB at the left node and pre-MB at
+            # the right node so the SIA VJP is linearized on the true forward trajectory.
+            H_itp_SIA = if simulation.parameters.simulation.use_MB
```
albangossard, line 314: I'm not sure you need such a complex implementation. We already have a similar mechanism based on the duplicity of time indices, for example in `effect_MB!`. I guess Claude will be able to correct me if my hypothesis in wrong :)

src/inverse/SIA2D/VJPs.jl
albangossard: Issue opened to investigate later: https://github.com/ODINN-SciML/Sleipnir.jl/issues/197

src/inverse/SIA2D/gradient.jl
albangossard: Issue opened to investigate later: https://github.com/ODINN-SciML/ODINN.jl/issues/532

albangossard (commented):
@JordiBolibar I made the different releases and now we need to merge this PR with https://github.com/ODINN-SciML/ODINN.jl/pull/527. I'll merge everything in `dev`, so I opened issues to keep track of my suggestions (see below).

Issue #493, opened 2026-05-15
Development tools are listed as runtime dependencies


haakon-e:
All four Julia packages list pure development tools under `[deps]` rather than `[extras]`/`[targets]`:

- `Sleipnir.jl/Project.toml`: has `Revise`, `Infiltrator`, `BenchmarkTools`, `JET` in `[deps]`. The `[extras]` block only contains `Aqua` and `Test`.
- `Huginn.jl/Project.toml`: same: `Revise`, `Infiltrator`, `BenchmarkTools` in `[deps]`.
- `Muninn.jl/Project.toml`: same.
- `ODINN.jl/Project.toml`: same.

The four packages also each do `using Infiltrator` in their top-level module file.

Suggested changes:
- Remove `Revise`, `Infiltrator`, `BenchmarkTools`, `JET` from every `[deps]` block. Move them to `[extras]` and add them to the relevant targets (e.g. `test`, `bench`).
- Drop the `using Infiltrator` from each top-level module file. The debugger should be loaded interactively (`using Infiltrator` in the REPL) by developers who need it, not by users.

---

xref: https://github.com/openjournals/joss-reviews/issues/10528

haakon-e:
Another imported package in `ODINN.jl` that seems to be unused (and thus should likely be removed) is `TensorBoardLogger`.

JordiBolibar:
Addressed in #536

PR #536, opened 2026-08-30
Misc changes for JOSS review

### LINKED ISSUE #487: Instantiation warnings
Instantiating the package locally on Julia v1.12.6 I get the following output:

```Julia
Precompiling packages finished.
  570 dependencies successfully precompiled in 221 seconds. 383 already precompiled.
  5 dependencies had output during precompilation:
┌ DateFormats
│  WARNING: Constructor for type "DateTime" was extended in `DateFormats` without explicit qualification or import.
│    NOTE: Assumed "DateTime" refers to `Dates.DateTime`. This behavior is deprecated and may differ in future versions.`
│    NOTE: This behavior may have differed in Julia versions prior to 1.12.
│    Hint: If you intended to create a new generic function of the same name, use `function DateTime end`.
│    Hint: To silence the warning, qualify `DateTime` as `Dates.DateTime` in the method signature or explicitly `import Dates: DateTime`.
└  
┌ Muninn
│  [ Info: Downloading preprocessed data
│  [ Info: Downloading RGI stats from Bremen cluster
└  
┌ WebSockets
│  WARNING: Imported binding Logging.termlength was undeclared at import time during import to WebSockets.
│  WARNING: Imported binding Logging.showvalue was undeclared at import time during import to WebSockets.
└  
┌ ODINN
│  ┌ Warning: ODINN requires Julia 1.10 or 1.11. You are using Julia 1.12.6, which is not supported.
│  └ @ ODINN ~/Documents/reviews/JOSS/ODINN.jl/src/ODINN.jl:20
└  
┌ Cassette
│  ┌ Warning: Cassette.jl may not work on this version of Julia
│  │   Base.VERSION = v"1.12.6"
│  └ @ Cassette ~/.julia/packages/Cassette/ROd5e/src/overdub.jl:600
└  
```

It is not clear to me where the `DateTime` and `Logging` warnings originate from, but these would be good to resolve as they may slow or inhibit precompilation as far as I understand.

The docs also state that

> `ODINN.jl` and the rest of the ODINN ecosystem packages require Julia v1.11 or later

If 1.12 is not supported, it would be useful to clarify this in the docs.

---

xref: https://github.com/openjournals/joss-reviews/issues/10528

### LINKED ISSUE #494: Every package does `cd(@__DIR__)` at module load
Every top-level module (ODINN, Muninn, Huginn, Sleipnir) calls `cd(@__DIR__)` in their `src/PackageName.jl` file. This changes the user's current working directory as a side-effect of `using PackageName`. It _appears_ this is done solely to set
```julia
const global root_dir::String = dirname(Base.current_project())
```
which is then used to include the containing files
A more canonical approach would be to simply do
```julia
include("setup/config.jl")
# Losses
include("losses/Losses.jl")
include("losses/Regularization.jl")
include("losses/MultiLoss.jl")
# All parameters needed for the models
include("inverse/VJPTypes.jl")
include("inverse/AdjointTypes.jl")
include("parameters/Hyperparameters.jl")
include("parameters/UDEparameters.jl")
# etc...
```
This way, users running simulations from another environment that has e.g. `ODINN.jl` as a dependency, won't see their working directory change as a result of loading the package.

The few other instances using e.g. `ODINN.root_dir` to reference a path, can instead use `pkgdir(ODINN)`.

---

xref: https://github.com/openjournals/joss-reviews/issues/10528

albangossard:
## Main changes

- Remove `cd(@__DIR__)` at module load #494 
- Fix precompilation by moving the definition of `Model(; ...)` to Sleipnir #487 
- Clarify supported versions of Julia and drop Julia 1.10 support #487 
- Remove development tools from runtime dependencies https://github.com/ODINN-SciML/ODINN.jl/issues/493
- Add a mechanism to revise the workers depending on the active environment


src/setup/config.jl
```diff
@@ -120,7 +120,7 @@ function enable_multiprocessing(params::Sleipnir.Parameters)
                         redirect_stdout(devnull)
                         redirect_stderr(devnull)
                         using Dates
-                        @everywhere using Revise
+                        # @everywhere using Revise
```
JordiBolibar: Why is this changed? Doesn't this break multiprocessing?

JordiBolibar (approved):
Mostly a comment on multiprocessing.

src/setup/config.jl
albangossard: This was hard to make it work but in the end I came up with a solution that I think is satisfying 3b12bbe
By default when we run ODINN in the main environment where Revise is not accessible, there is no `@everywhere using Revise` in the workers. When we run in one of the three environments `test/test_env`, `docs`, or `scripts` Revise runs in the workers.
I checked the behaviour locally and this works well.

docs/src/glaciers.md
```diff
@@ -33,11 +33,24 @@ The main climate data supported include:
 
 ### Standard workflow
 
-In order to create `Glacier` types with information of a given glacier for a simulation, one can initialize a list of glaciers based on RGI (Randolph Glacier Inventory) IDs. Before running this, make sure to have used `Gungnir` to download all the necessary data for those glaciers in local. [Here](https://github.com/ODINN-SciML/Gungnir/blob/main/notebooks/Example.ipynb) you will find a notebook showing how to do so. Even easier, you can just double check that these glaciers are already available on the ODINN server. The list of the already processed glaciers can be obtained with `get_rgi_paths()`.
+Alpine glaciers are identified based on RGI (Randolph Glacier Inventory) IDs. There are two options to create `Glacier` types containing information about a given glacier for a simulation:
```
JordiBolibar: Why Alpine glaciers? I would make this more generic.

JordiBolibar (approved):
Just a small change on the RGI.

PR #544, opened 2026-09-07
Use var-specific bounds in GlacierWideInv

### LINKED ISSUE #514: `GlacierWideInv` applies A bounds to all parameters
`GlacierWideInv` uses `minA`/`maxA` from `PhysicalParameters` for its atanh re-parameterization unconditionally, regardless of which physical variable is passed as var. This means `GlacierWideInv(params, glaciers, :C)`applies the wrong bounds, producing an incorrectly scaled initial `θ` for C inversions. The constructor should accept a var-specific bounds pair, or look up `min{var}`/`max{var}` from `PhysicalParameters`.

@girodlis We would need this for your C gridded inversions, otherwise you might struggle to invert the C fields, since they will be constrained by the A value bounds.

JordiBolibar:
`GlacierWideInv` applied the `minA`/`maxA` atanh re-parameterization regardless of which variable was passed as `var`, so `GlacierWideInv(params, glaciers, :C)` built a θ₀ scaled by the A bounds. This turned out to be worse than a scaling error: `Glacier2D.C` defaults to 0, so the constructor evaluated `atanh(2*(0 - minA)/(maxA - minA) - 1)` and threw a `DomainError`.

Worth flagging one subtlety, since the issue suggested looking up `min{var}`/`max{var}` from `PhysicalParameters`: that would still be wrong for C. `LawC` applies `C = maxC * (tanh(θ) + 1) / 2`, whose lower bound is 0, not `minC`. The θ₀ has to invert the map the law actually applies, so the bounds pair for `:C` is `[0, maxC]` and `minC` stays unused here. `GriddedInv` had already worked this out, so this mirrors it rather than introducing a second convention — the two containers feed the same laws and must agree.

`C = 0` has no finite preimage and is now seeded just inside the bound. Unsupported variables error instead of silently producing nonsense, matching `GriddedInv`. `minval`/`maxval` overrides are available for callers who want non-default bounds.

Note this is currently masked by Sleipnir's defaults, where `maxA == maxC == 8e-17` — it only bites once a realistic `maxC` is set, which is still an open TODO there.

@girodlis this should unblock the C gridded inversions.

## Testing

Adds `test_C_parameterization()`, which had no equivalent before — `LawC` appears nowhere in the suite, and the existing "C>0" adjoint tests only assign `glacier.C` as a constant, never going through the inversion containers. It checks that `:C` round-trips through `LawC`'s forward map for both containers, that `C = 0` stays finite, and that unsupported vars error. No downloads and no PDE solve; runs in 3.9 s.

Verified it fails against the old code, where the first assertion feeds `atanh(74.07)`.

Fixes #514

### DIFF
--- a/test/inversion_test.jl
+++ b/test/inversion_test.jl
@@ -271,3 +271,43 @@ function inversion_test(;
         @test minimum(rel_error) < 1e-4
     end
 end
+
+"""
+    test_C_parameterization()
+
+Check that the θ₀ built for `:C` inverts the tanh parameterization applied by `LawC`.
+`maxC` is set well apart from the A bounds so that applying the A bounds to C is caught.
+The `:A` round-trip is already covered by the scalar and gridded inversion tests above.
+"""
+function test_C_parameterization()
+    maxC = 1e-15
+    C = 3e-16
+    params = Parameters(
+        simulation = SimulationParameters(
+            tspan = (2010.0, 2012.0), multiprocessing = false,
+            use_MB = false, test_mode = true, working_dir = Huginn.root_dir),
+        physical = PhysicalParameters(minA = 8e-21, maxA = 8e-18, maxC = maxC)
+    )
+
+    nx, ny = 5, 5
+    toy_glacier(C) = Glacier2D(
+        rgi_id = "toy", H₀ = ones(nx, ny), S = ones(nx, ny), B = zeros(nx, ny),
+        A = 5e-19, C = C, n = 3.0, Δx = 50.0, Δy = 50.0, nx = nx, ny = ny)
+
+    forwardC(θ) = maxC * (tanh(θ) + 1) / 2 # The map applied by LawC
+    glaciers = Vector{Sleipnir.AbstractGlacier}([toy_glacier(C)])
+
+    @test all(forwardC.(GlacierWideInv(params, glaciers, :C).θ.θ[Symbol("1")]) .≈ C)
+    @test all(forwardC.(GriddedInv(params, glaciers, :C).θ.θ[Symbol("1")]) .≈ C)
+
+    # C = 0 is the Glacier2D default and has no finite preimage, so it must be seeded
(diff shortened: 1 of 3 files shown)


test/inversion_test.jl
```diff
@@ -271,3 +271,43 @@ function inversion_test(;
...
+
+    @test all(forwardC.(GlacierWideInv(params, glaciers, :C).θ.θ[Symbol("1")]) .≈ C)
+    @test all(forwardC.(GriddedInv(params, glaciers, :C).θ.θ[Symbol("1")]) .≈ C)
+
+    # C = 0 is the Glacier2D default and has no finite preimage, so it must be seeded
+    # rather than sent to atanh(-1) = -Inf
+    zero_C = Vector{Sleipnir.AbstractGlacier}([toy_glacier(0.0)])
+    for θ in (GlacierWideInv(params, zero_C, :C).θ, GriddedInv(params, zero_C, :C).θ)
+        @test all(isfinite, θ.θ[Symbol("1")])
+        @test all(forwardC.(θ.θ[Symbol("1")]) .< 1e-3 * maxC)
```
albangossard: This test will break as soon as `maxC` becomes too small. Since that edge case is handled using `Sleipnir.Float(-5)` in `GlacierWideInv`, this test shouldn't use `maxC` but do something like 
```suggestion
        @test all(forwardC.(θ.θ[Symbol("1")]) .<= forwardC(Sleipnir.Float(-5))
```

albangossard (approved):
One small comment, otherwise LGTM
By the way could you create an issue in Sleipnir that mentions we should remove `minC` please?

JordiBolibar:
> One small comment, otherwise LGTM By the way could you create an issue in Sleipnir that mentions we should remove `minC` please?

Done! https://github.com/ODINN-SciML/Sleipnir.jl/issues/202#event-30747380743

PR #545, opened 2026-09-08
Switch docs build from main to dev

JordiBolibar:
Make `dev` the default to build the docs, this way we can easily integrate the latest changes directly into the documentation.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -3,7 +3,7 @@ name: Documentation
 on:
   push:
     branches:
-      - main # update to match your development branch (master, main, dev, trunk, ...)
+      - dev # update to match your development branch (master, main, dev, trunk, ...)
     tags: '*'
     paths-ignore:
       - 'test/**'
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -3,7 +3,8 @@ name: Documentation
 on:
   push:
     branches:
-      - dev # update to match your development branch (master, main, dev, trunk, ...)
+      - dev
+      - main
     tags: '*'
     paths-ignore:
       - 'test/**'
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -134,10 +134,14 @@ makedocs(
 )
 
 if get(ENV, "CI", nothing)=="true"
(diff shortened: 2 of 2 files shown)


.github/workflows/documentation.yml
```diff
@@ -3,7 +3,7 @@ name: Documentation
 on:
   push:
     branches:
-      - main # update to match your development branch (master, main, dev, trunk, ...)
+      - dev # update to match your development branch (master, main, dev, trunk, ...)
```
albangossard: In line of my global comment, I suggest that we set both `dev` and `main`

albangossard (changes requested):
Isn't this already the case?
image

I recommend that we keep `main` as the default branch: imagine that we develop a huge breaking change that affects the documentation. During the time that lives on `dev` and not on `main`, all the users will see the new documentation instead of the one that corresponds to the version they are using.

There is probably a way to make `main` the default branch while still being able to select the `dev` branch on the bottom (see screenshot above).

.github/workflows/documentation.yml
JordiBolibar: I agree with this solution, the only problem for this is that it implies building the docs twice, which means even more CI time... I'm not that worried that there will be massive misalignments between `main` and `dev`.  And if we are honest, most of the userbase will work directly on `dev`, since it's mostly us around who are developing the model.

I would say that if we ever see increased model adoption, then it might make sense to switch to main, if people are just users. But for now all of us are devs, so it makes more sense to me to be able to directly display the `dev` docs. Happy to hear your thoughts on this :)

albangossard: I thought that the various builds could come from different triggers of the CI. But yes if this implies building the docs twice, then I agree that it's best to keep it simple.

albangossard (approved):

.github/workflows/documentation.yml
JordiBolibar: I'll investigate that, and if it can be done using independent triggers, then I'll implement it, otherwise I'll stick to `dev`.

JordiBolibar: I implemented an independent trigger for `dev` and `main`. Since we're pushing to `dev` here,  this should add the `dev` version on the docs. Once merged, I'll see if both versions are available on the public docs.

PR #547, opened 2026-09-11
Differentiate mass balance in the RHS, delete the jump adjoint

JordiBolibar:
Mass balance is now evaluated as a **continuous source term** in the ice flow right-hand side instead of a periodic callback.

This enables **automatic adjoints** (SciMLSensitivity) to differentiate through mass balance, which was previously impossible. The adjoint now includes the mass balance contribution `λ_∂ṁ∂H` (elementwise, as ∂ṁ/∂H is diagonal). Removed 78 lines from `gradient.jl` that manually stitched the MB jump into the adjoint.

---

### In CI: `Core12`
5 regression tests (~5 min) for the **MB-dominated gradient regime**. These shrink the ice-flow parameter `A` so the gradient is dominated by mass balance, specifically testing the new adjoint term. Existing tests (Core1–Core11) are ice-flow-dominated and would miss MB adjoint errors.

### Not in CI
   Group | Purpose | Why |
 |-------|---------|-----|
 | `Core12teeth` | Expected to **fail** | Sanity check: if it passed, the new adjoint would be identical to the old callback-based one |
 | `Core12fd` | Sweeps finite difference step sizes | Diagnostic, not a regression guard |
 | `GradTolGrid`, `GradTol` | Checks solver tolerance stability | Diagnostic |

---
## Model interface

Mass balance models must now implement:
- `MB_rate!`: Rate in m/yr, evaluated at every solver step
- `MB_rate_∂H!`: Diagonal Jacobian ∂ṁ/∂H
- `init_mb_cache`: Precompute climate per mass balance window

See [docs/src/extending.md](docs/src/extending.md) for details.

---
## Known limitations

- Only `TImodel1` (`mb_S_dependence = :elevation_only`) works today. The `:general` case is a stub; tracked in [Muninn.jl#72](https://github.com/ODINN-SciML/Muninn.jl/issues/72).
- `CustomMLP` forward script does not run yet.

Companion PRs:
- https://github.com/ODINN-SciML/Sleipnir.jl/pull/203
- https://github.com/ODINN-SciML/Huginn.jl/pull/180
- https://github.com/ODINN-SciML/Muninn.jl/pull/73

### DIFF
--- a/docs/src/extending.md
+++ b/docs/src/extending.md
@@ -134,9 +134,9 @@ New mass balance models subtype `MBmodel` (defined in `Muninn`). Mass balance is
 \frac{\partial H}{\partial t} = -\nabla\cdot(D\nabla S) + \dot m(H, t)
 ```
 
-so a model supplies a **rate**, evaluated at every step of the solve, rather than an increment applied by a callback. There is no mass balance callback: that is what lets the automatic adjoint differentiate through mass balance, and it removes the operator splitting error a periodic jump introduces.
+so a model supplies a **rate**, evaluated at every step of the solve — not an increment applied by a callback. That's what lets the automatic adjoint differentiate through mass balance, and it removes the operator-splitting error a periodic jump introduces.
 
-Evaluating a model inside the right hand side is not free. It runs orders of magnitude more often than a monthly callback did, it may not read the climate rasters (`Rasters` is not differentiable and not cheap), and it has to be differentiable with respect to `H`. The climate is therefore precomputed per mass balance window when the cache is built, and the model reads that.
+Evaluating a model inside the right hand side isn't free: it runs far more often than a monthly callback did, can't read the climate rasters (`Rasters` isn't differentiable and isn't cheap), and must be differentiable in `H`. So climate is precomputed per mass balance window when the cache is built, and the model reads that.
 
 ```
 Every RHS call:
--- a/docs/src/extending.md
+++ b/docs/src/extending.md
@@ -128,14 +128,24 @@ end
 
 ## Add a new mass balance model
 
-New mass balance models subtype `MBmodel` (defined in `Muninn`). The MB callback fires every `step_MB` and runs three steps in sequence — only `compute_MB` is model-specific:
+New mass balance models subtype `MBmodel` (defined in `Muninn`). Mass balance is a source term of the ice flow right hand side,
+
+```math
+\frac{\partial H}{\partial t} = -\nabla\cdot(D\nabla S) + \dot m(H, t)
+```
+
+so a model supplies a **rate**, evaluated at every step of the solve, rather than an increment applied by a callback. There is no mass balance callback: that is what lets the automatic adjoint differentiate through mass balance, and it removes the operator splitting error a periodic jump introduces.
+
+Evaluating a model inside the right hand side is not free. It runs orders of magnitude more often than a monthly callback did, it may not read the climate rasters (`Rasters` is not differentiable and not cheap), and it has to be differentiable with respect to `H`. The climate is therefore precomputed per mass balance window when the cache is built, and the model reads that.
 
 ```
-MB callback (every step_MB):
-  ├── MB_timestep!(cache, model, glacier, step, t, glacier_idx)  # writes the MB into cache.iceflow.MB
-  │     └── compute_MB(mb_model, climate_2D_step, step)     ← implement this for your model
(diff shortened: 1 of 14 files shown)


docs/src/extending.md
```diff
@@ -156,6 +190,16 @@ function Muninn.compute_MB(model::MyMBmodel, climate_step::Climate2Dstep,
 end
 ```
 
+It is what `calibrate_MB_model` fits parameters against, and what the test suite checks `MB_rate!` agrees with. If the two ever drift apart, the model being calibrated stops being the model being integrated, which nothing else guards against.
+
+!!! note "Positivity is the model's responsibility"
+
+    `ṁ` must vanish as `H` approaches zero, or a cell melts through the bed. `TImodel1` does this with a cubic `smoothstep` ramp on the rate, which is C¹ with a bounded derivative — a hard mask would be a step discontinuity in the state that no adaptive error controller can resolve and no adjoint can differentiate. See `mass_balance_rhs.jl` in Muninn.
```
albangossard, line 197:
```suggestion
    `ṁ` must vanish as `H` approaches zero, or a cell melts through the bed. `TImodel1` does this with a cubic `smoothstep` ramp on the rate, which is C0 with a bounded derivative — a hard mask would be a step discontinuity in the state that no adaptive error controller can resolve and no adjoint can differentiate. See `mass_balance_rhs.jl` in Muninn.
```

docs/src/models.md
```diff
@@ -87,6 +87,10 @@ model = Model(
 
 `CustomMLP` is a subtype of `MBmodel` and wraps a `Lux.jl` feedforward network whose architecture, input feature normalisation bounds, and pre-trained weights are all read directly from the JSON export. The network takes monthly ERA5 climate features as inputs (e.g. `t2m`, `tp`, `ssrd`, …) and outputs a surface mass balance rate in m w.e. per time step. For now, only monthly time steps are supported. It is the *de facto* data-driven surface mass balance model in the ODINN ecosystem.
 
+!!! warning "Not yet usable in a simulation"
+
+    Mass balance is evaluated as a source term of the ice flow right hand side, and only models that declare `mb_S_dependence` as `:elevation_only` have that form today — in practice `TImodel1`. `CustomMLP` reads several fields, so it falls to the `:general` branch, which is not implemented: building a simulation with it and `use_MB = true` raises an error naming the functions to implement. See [Add a new mass balance model](@ref) for the interface.
```
albangossard, line 92: If I understand correctly, with the `CustomMLP` this won't be possible to build a LUT because we would have to build a table with dimensions (N_H, N_slope, N_aspect, N_time) which won't fit into memory. Am I right?
We would have to find a way to "tensorize" the structure of the MB model so that we can build input specific LUTs.

src/inverse/SIA2D/gradient.jl
```diff
@@ -127,14 +127,6 @@ function SIA2D_grad_batch!(θ, simulation::Inversion)
...
 
         if typeof(simulation.parameters.UDE.grad) <: DiscreteAdjoint
-            tstopsMB = if simulation.parameters.simulation.use_MB
-                tstopsMB = Huginn.define_callback_steps(tspan, simulation.parameters.simulation.step_MB)[2:end] # Discard first time step to be aligned with the forward
-                @assert all(map(ti -> ti in t, tstopsMB)) "When using the DiscreteAdjoint the tstops of the MB callback must all be included in the tstops from the results."
-                tstopsMB
-            else
-                []
-            end
-
```
albangossard, line 137: Happy to see all of this disappear!

test/gradient_diagnostics.jl
albangossard, line 1: What is the cost of these tests? Because if it's not covered by the CI, then we have no guarantee this won't break in the future.

test/test_grad_loss.jl
```diff
@@ -153,8 +184,15 @@ function test_grad_finite_diff(
...
             step = δt,
+            abstol = abstol,
+            adaptive = adaptive,
+            dt = dt,
             progress = true,
-            solver = useSciMLSenseAlg ? ROCK4() : RDPK3Sp35() # Use another solver when using SciMLSensitivity because `InterpolatingAdjoint` is not stable with our ODE in backward mode
+            solver = if !isnothing(solver)
+                solver
+            else
+                useSciMLSenseAlg ? ROCK4() : RDPK3Sp35() # Use another solver when using SciMLSensitivity because `InterpolatingAdjoint` is not stable with our ODE in backward mode
```
albangossard, line 194: Could you try to remove this and use the default solver? I suspect this was the MB callback that caused the unstable backward ODE.

